Issue #5, opened 2017-10-18
Introduce models


arthurdm:
We need to introduce OAS models (as interfaces) so that we can work with the OAS tree during annotation processing.  These models should be able to be chained, as well as created independently.  

We also need a factory to create these.

arthurdm:
We should also include a sample on how to build a model tree.

arthurdm:
Here's a sample that uses the swagger libraries https://github.com/microservices-api/oas3-configuration/blob/master/sample.oas.programmingModels/src/io/swagger/sample/config/OpenAPIConfigurationImpl.java#L23

We're trying to achieve a similar pattern here, but using MP Config as the entry point of this configuration.

arthurdm:
Models have been introduced in the `api` section.  I will open a new issue for the documentation of a sample usage of the models via the model reader.

PR #17, opened 2017-10-24
MicroProfile OpenAPI Models

arthurdm:
Initial drop for https://github.com/eclipse/microprofile-open-api/issues/5

This set of models will allow application devs to build an OpenAPI v3 tree programmatically (either a complete tree, or to augment or filter the annotations).

This initial drop introduces the models, while subsequent PRs will introduce filtering, model configuration, etc.

Consider 3 scenarios where these models will be used:
(1) The app developer wants to programmatically create an OpenAPI v3 document (ie: think of a DOM tree) instead of using annotations or pre-generating a YAML document.

(2) The app developer wants to build a "starter model" that will be used as the base OAS3 tree populated during annotation processing.  For example: they might use this approach to pre-populate the leading elements such as Info, License, etc, to keep it separated from the REST models.

(3) The app developer wants to filter or augment certain parts of the OpenAPI application dynamically depending on the environment.  They can setup a filter (upcoming PR) that will inject the current model tree and allow for filtering / augmentation.


arthurdm:
This PR is under progress, as we fix the javadocs.

arthurdm:
Please note that the failed ip-validation is a bug in eclipse's validation tool, which is incorrectly flagging the PR commits from `noreply@github.com` (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=526426)

arthurdm:
@Emily-Jiang - can you please take a look into this PR?   Thanks.

Emily-Jiang:
@arthurdm I started last night and will continue tomorrow. It is a fairly big PR.

arthurdm:
Thanks Emily.  As we discussed, this sample helps visualize the programming model that we're going to use the models for:  https://github.com/microservices-api/oas3-configuration/blob/master/sample.oas.programmingModels/src/io/swagger/sample/config/OpenAPIConfigurationImpl.java#L23

EricWittmann:
This PR looks good to me.  I'm not sure if there is a policy in place regarding squashing commits for a PR, but my suggestion would be to squash all the commits into a single commit for each accepted PR.  It cleans up the codebase a lot, but at a loss of some detail/fidelity.  Probably personal preference. :)

Otherwise I give this a 👍 and 🎆

Emily-Jiang:
Merged this. Though the PR has a warning of complaining ip validation, Arthur said this is an Eclipse bug and further to the discussion on the  mailing list, this should not block the merging.

Issue #24, opened 2017-11-10
Callback model should implement Reference interface


arturdzm:
Example of $ref to callback

```
openapi: '3.0.0'
info:
  title: API
  version: 1.0.0

paths:
  /path:
    get:
      description: 'Path with callback'
      responses:
        200:
          description: 'ok'
      callbacks:
        simpleCallback:
          $ref: '#/components/callbacks/simpleCallback'

components:
  callbacks:
    simpleCallback:
      '/mycallback':
        get:
          responses:
            200:
              description: 'ok'
```

arthurdm:
Issue has been resolved by pull request #25

PR #25, opened 2017-11-10
Fix for issue #24 - Add Reference interface to Callback model

arturdzm:
Added a Reference extension to Callback model class


arthurdm:
Looks good, thanks @arturdzm

PR #32, opened 2017-11-16
Model filter and reader

arthurdm:
-  The OASModelReader is the way that users contribute a model tree:  either a complete tree + disabled annotation, or stub tree + (annotation | pre-generated file)    

- The OASFilter is a general filter, which can filter model elements anytime the vendor creates new models, which is during two cases:  (1) processing annotations, or  (2) parsing a document (either complete or stub).   The last method "void filterOpenAPI" is the last called method of the filter, and returns void because we don't want to give users the ability to not return at least an OpenAPI (but they remove any of the other elements by returning null to any of the filter methods).

- Changed `ApiResponse` to `APIResponse` and `ApiResponses` to `APIResponses`.  It was weird to have `Api`, since we always refer to it with capital letters.

An app developer would register their filter / reader with the following keys:

```
mp.openapi.filter=myPackage.myFilter
mp.openapi.model.reader=myPackage.myReader
```

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/OASReader.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/OASReader.java
@@ -32,5 +32,5 @@ public interface OASReader {
      * 
      * @return the OpenAPI model to be used by the vendor
      */
-	OpenAPI read();
+    OpenAPI read();
 }
--- a/api/src/main/java/org/eclipse/microprofile/openapi/OASReader.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/OASReader.java
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


arthurdm:
@leochr - can you please review this PR when you get a chance?

api/src/main/java/org/eclipse/microprofile/openapi/OASReader.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.eclipse.microprofile.openapi.models.OpenAPI;
+
+/**
+ * This interface allows application developers to programmatically contribute an OpenAPI model tree.  
+ * 
+ * In this scenario the developer can choose whether to provide the entire OpenAPI model while disabling annotation
+ * scanning, or they can provide a starting OpenAPI model to be augmented with the application annotations.  
+ *
+ */
```
leochr: Should the registration process be documented here (similar to the documentation for the filter)?

api/src/main/java/org/eclipse/microprofile/openapi/OASReader.java
```diff
@@ -0,0 +1,36 @@
...
+public interface OASReader {
+
+    /**
+     * This method is called by the vendor's OpenAPI processing framework.  It can be a fully complete and valid OpenAPI
+     * model tree, or a partial base model tree that will be augment by either annotations or pre-generated OpenAPI documents.
```
leochr: A minor correction is needed here:
"will be **augment** by either" --> "will be **augmented** by either"

arthurdm:
thanks for the feedback!  all the review comments have been addressed in the last commit.

Issue #3, opened 2017-10-16
Add annotation support


arthurdm:
This issue will add OpenAPI v3 annotations that allow users to document their RESTful application.  It should cover the entire spectrum of the OpenAPI v3 specification

arthurdm:
Closing this issue as there has been many PRs that delivered this annotations in the past few weeks.  We consider the current set of annotations very close to complete, with just a minor few bugs/enhancements, which will be tracked via individual issues.

PR #39, opened 2017-11-16
Annotation Javadocs updated

a-saf:
This PR is for issue #3


api/src/main/java/org/eclipse/microprofile/openapi/annotations/Operation.java
```diff
@@ -45,21 +45,24 @@
...
      * Tags can be used for logical grouping of operations by resources or any other qualifier.
-     *
+     * </p>
      * @return the list of tags associated with this operation
      **/
     String[] tags() default {};
 
     /**
-     * Provides a brief description of this operation.
+     * Provides a brief description of this operation does.
```
arthurdm: Maybe it should be: Provides a brief description of `what` this operation does.

Issue #41, opened 2017-11-20
Specific an opinion on internationalization


arthurdm:
We don't need to mandate a solution, but the spec should at least mention this in the limitations section

arthurdm:
Resolved with merge #42

PR #42, opened 2017-11-20
Updating limitations with internationalization

arthurdm:
PR for https://github.com/eclipse/microprofile-open-api/issues/41

For 1.0 it's probably premature to come up with a full internationalization framework, so I propose that we call this out as a limitation but suggest a reasonable approach for vendors that do want to support it.

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
@@ -45,7 +45,7 @@ This document represents the result of the applied <<Processing rules>>.
 The default `Content-Type` of the `/openapi` endpoint is `application/yaml`.
 
 Vendors must also support the type `application/json` if requested via the
-`Content` header.
+`Accept` header.
 
 === Query parameters
 No query parameter are required for the `/openapi` endpoint.  However, one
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/microprofile-openapi-spec.adoc
```diff
@@ -45,7 +45,7 @@ This document represents the result of the applied <<Processing rules>>.
 The default `Content-Type` of the `/openapi` endpoint is `application/yaml`.
 
 Vendors must also support the type `application/json` if requested via the
-`Content` header.
+`Accept` header.
```
EricWittmann: +1 to this bug fix. :)

spec/src/main/asciidoc/microprofile-openapi-spec.adoc
```diff
@@ -135,3 +135,12 @@ overriding any conflicting elements from the current model.
...
 == Limitations
+
+=== Internationalization
+The 1.0 version of the MicroProfile OpenAPI spec does not require vendors to
+support multiple languages based on the `Accept-Language`.  One reasonable
+approach is for vendors to support unique keys (instead of hardcoded text) via
+the various <<Documentation Mechanisms>>, so that the implementing framework can
+perform a global replacement of the keys with the language-specific text that
+matches the `Accept-Language` request for the `/openapi` endpoint.  A cache of
+processed languages can be kept to improve performance.
```
EricWittmann: This seems reasonable as long as it's consistent with the approach taken by other MP specs.

arthurdm:
I checked with Emily and none of the other MP specs had to deal with internationalization yet.   So I think we have a reasonable limitation with a very good suggestion for vendors that want to go ahead and implement it.

PR #43, opened 2017-11-20
Components to hold reusable items + references from other annotations

leochr:
Enhance annotations to support definition of Components items and allow referring them from other annotation such as Parameter, RequestBody, etc


arthurdm:
looks great, thanks @leochr !

PR #44, opened 2017-11-21
Fixed some minor spelling/typos

EricWittmann:
I found some minor errors while reviewing the spec.


arthurdm:
Good catch @EricWittmann !     btw - there's an `ip-validation` that fails if we don't use `-s` to sign the Git commits - it doesn't block the PR, just fyi.

EricWittmann:
Oh OK no problem.  I'll sign PRs from now on.  <<goes off to Google how to sign a PR>>

PR #58, opened 2017-11-24
OASFactory tests for invalid input. Each test expects an exception.

mrglavas:
Signed-off-by: Michael Glavassevich <mrglavas@ca.ibm.com>


mrglavas:
Note that the method signature for OASFactory.createObject() is "<T extends Constructible> T createObject(Class<T> clazz)". This blocks any Class object that isn't a Constructible at compile time, so haven't included any tests for that since they won't compile anyway.

arthurdm:
Thanks Michael, looks good to me.

PR #70, opened 2017-11-27
Additional configurations plus document exception.

leochr:


leochr:
@arthurdm please review.

I'll push the example values for configs separately as I am hitting some formatting issues.

PR #73, opened 2017-11-27
Moved the filter step to the end of model processing

EricWittmann:
The filtering should be done once after all other model processing is complete.


EricWittmann:
All checks have passed!  Woo hoo!

PR #75, opened 2017-11-28
Add example values for configs

leochr:
Signed-off-by: leochr <leojc@ca.ibm.com>

Added sample values for config properties.
Some additional descriptions for config properties.


arthurdm:
thanks for these samples Leo!

PR #77, opened 2017-11-28
Fixes to ensure correct sub-types are returned from Parameter builder methods.

mrglavas:
Signed-off-by: Michael Glavassevich <mrglavas@ca.ibm.com>


mrglavas:
This PR updates the Parameter interface to include a type parameter which is returned from each of the builder methods. This ensures that the correct sub-type (e.g. QueryParameter) is always returned from the builder method. Also made the "in" property read-only.

PR #78, opened 2017-11-28
Changes to model: Components

EricWittmann:
Modified add** methods to no longer be plural (since they all add only one item at a time).


EricWittmann:
I think this `add***()` pattern is repeated throughout.  So expect more PRs for each model as I continue to work on the WF Swarm impl. :)

PR #79, opened 2017-11-28
Changes to model: OpenAPI

EricWittmann:
Added missing "add" methods passing through to the Components property.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/models/OpenAPI.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/models/OpenAPI.java
@@ -19,16 +19,8 @@ package org.eclipse.microprofile.openapi.models;
 
 import java.util.List;
 
-import org.eclipse.microprofile.openapi.models.callbacks.Callback;
-import org.eclipse.microprofile.openapi.models.examples.Example;
-import org.eclipse.microprofile.openapi.models.headers.Header;
 import org.eclipse.microprofile.openapi.models.info.Info;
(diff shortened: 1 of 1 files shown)


arthurdm:
hey @EricWittmann - the direct members of the OpenAPI model map to the OpenAPI element in the spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#oasObject

It seems the fields you're adding in this PR are already present in the Components model, which maps to the Component level of the spec https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#componentsObject

Do you want to update this PR to only have the updates to the method names, such as `addServersItem` -> `addServer`?

EricWittmann:
@arthurdm Right I only added those methods because the OpenAPI model already has the following two methods:

https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/models/OpenAPI.java#L248-L265

I think we either have to add the ones included in my PR or remove the two that I've linked above.  Honestly I think removing the two linked above makes the most sense.  I would have done that but I assumed there was a reason why those two were included.

I'm cool either way. :)

arthurdm:
ah, now I see what you mean. =)    

yes, definitely agree that we should be consistent.  I propose that we remove those 2 extra methods, so that we're closer to the spec.

EricWittmann:
+1 I'll update the PR tomorrow.

EricWittmann:
@arthurdm OK I've updated this PR.  Should be more reasonable now. :)

PR #85, opened 2017-11-29
Allowing factory to bet set via SPI

arthurdm:
Allow the OASFactory to change after the class was loaded - in case there were timing issues in an OSGi or equivalent framework.


arthurdm:
@leochr - can you please take a look?

leochr:
Looks fine @arthurdm. Thanks.

PR #86, opened 2017-11-29
Add coverage for the Schema, Link, Example, ...

turkeylurkey:
ExternalDocumentation and Extension annotations

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
@@ -189,9 +189,7 @@ public class ReviewResource {
             description = "username of the user for the reviews",
             required = true,
             in = ParameterIn.PATH,
-            content = @Content(
-                examples = @ExampleObject(
-                    value = "bsmith")))
+            examples = { "bsmith", "pat@example.com" } )
         @PathParam("user") String user){
 
             List<Review> reviewsByUser = new ArrayList<Review>();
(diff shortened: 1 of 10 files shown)


tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
```diff
@@ -189,9 +189,7 @@ public Response getReviewByUser(
             description = "username of the user for the reviews",
             required = true,
             in = ParameterIn.PATH,
-            content = @Content(
-                examples = @ExampleObject(
-                    value = "bsmith")))
```
arthurdm: can we keep the current `content` field as well?    That way we're not decreasing coverage on other annotations.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
```diff
@@ -222,9 +220,7 @@ public Response getReviewByUser(
                 description = "name of the airlines for the reviews",
                 required = true,
                 in = ParameterIn.PATH,
-                content = @Content(
-                    examples = @ExampleObject(
-                        value = "Acme Air")))
```
arthurdm: same thing here - I think we can keep the `content`.

arthurdm:
thanks for PR @turkeylurkey - I just made 2 small comments, if you can please address those.

turkeylurkey:
Ok, I restored those @Contents.

PR #90, opened 2017-11-30
Adding validation notice on spec

arthurdm:
Adding a small note to the spec to say that validation is vendor-specific.  


arthurdm:
@EricWittmann - let me know if this makes sense.

PR #95, opened 2017-11-30
Annotation changes - added name attribute where needed + javadoc to clarify when attributes are required

leochr:
- Added name attribute to APIResponse and RequestBody so they can be stored under components
- Added javadoc to clarify when the name attribute is required and how it's used
- Added default values for all attributes of referenceable items, so user can easily define a reference by only specifying the 'ref' attribute
-  Clarified that for references other attributes are not required
- Added empty value to enums SecuritySchemeIn and SecuritySchemeType  and set it as default value (similar to ParameterIn and ParameterStyle).


leochr:
@arthurdm please review. thanks.

PR #97, opened 2017-12-01
Collapse Schema subtypes into one Schema model

arturdzm:
- Collapse all child Schemas into Schema interface
- Change Schema type() to SchemaType enum with default value of null for schemas with $ref


arthurdm:
thanks @arturdzm - in a future PR we'll need to correct a few javadocs that still says `ArraySchema` and `ComposedSchema`, but I will merge this PR now to avoid merge conflicts.

PR #109, opened 2017-12-05
Updating expected output for TCK

arthurdm:
Updating expected output for the TCK


a-saf:
Reviewed: looks good. With the latest change to apps in PR #114 , the info annotation will have slightly different values of attributes, but everything else is good.

arthurdm:
Thanks @a-saf - we'll update the output in the next generation from the impl.

PR #111, opened 2017-12-05
Model change: Schema#additionalProperties

### LINKED ISSUE #100: additionalProperties only supports Schema
According to the spec the value of the `additionalProperties` is:

> Value can be boolean or object. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema.

Our model currently only allows Schema

EricWittmann:
Support both Schema and Boolean for Schema#additionalProperties.  Fixes #100

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
@@ -33,6 +33,9 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 /**
  * Describes the Components object that holds various reusable objects for different aspects of the OpenAPI Specification (OAS).
+ * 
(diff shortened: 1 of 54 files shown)


EricWittmann:
@arthurdm This is a proposal - tell me what you think about having multiple setters to capture the intent of the `additionalProperties` property.  I think this makes it more explicit than simply having one set of `Object` setters.  But happy to do the latter if that's the consensus opinion.

arthurdm:
Thanks @EricWittmann - looks good.  Just two comments:  

(1) can we say that the 2 overloaded methods are mutual exclusive?  

(2) I wonder if using the word "validate" in the javadoc implies that vendors should validate and throw errors against this scenario (which we called out as not being required from the spec).

EricWittmann:
Good points.  I'll correct and update the PR.

EricWittmann:
It seems I have messed up the merge.  Closing this PR and I will submit a new one. :(

PR #122, opened 2017-12-06
Updating references to old types

arthurdm:
Updating the javadocs for `Schema` model.   It had older references to ArraySchema and ComposedSchema, as well as inconsistent case (schema vs Schema).  


navidsh:
Looks good @arthurdm

PR #127, opened 2017-12-07
Updating artifacts

arthurdm:
Updating a few artifacts that may be affecting the release build.


arthurdm:
Cancelling this PR as the rc1 build was successful.

PR #142, opened 2017-12-09
Update spec

arthurdm:
Updating spec to include missing items (when compared with other MP projects).


arthurdm:
@leochr - can you please take a look into this one?

leochr:
@arthurdm looks good, thanks!

PR #143, opened 2017-12-11
Minor suggested edits for RC2.

EricWittmann:

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
@@ -332,10 +332,10 @@ post:
     requestBody:
       description: Created user object
       content:
-       '*/*':
-        schema:
-          $ref: '#/components/schemas/User'
-          required: true
+        '*/*':
+          schema:
+            $ref: '#/components/schemas/User'
+            required: true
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/microprofile-openapi-spec.adoc
```diff
@@ -669,11 +669,12 @@ support the `https` protocol as well, to enable a secure connection to the OpenA
 endpoint.
 
 === Content format
-The default format of the `/openapi` endpoint is `YAML`.
+The default format of the `/openapi` endpoint is `YAML`, which should result in
+a response with a `Content-Type` header with a value of `application/yaml`.
```
arthurdm: There's an interesting behavior from browsers that trigger the "save as" wizard (instead of just displaying the file as text) if you return something with Content-Type `application/yaml`.  So I think we can leave this requirement out of the spec so that vendors can choose if they want to server YAML as a text (probably the preferred result), or if they want to server is a file you must save.   This is likely due to `application/yaml` not being an official MIME type as per https://www.iana.org/assignments/media-types/media-types.xhtml

EricWittmann:
Fair point.  Corrected. :)

PR #154, opened 2017-12-12
Adding missing POM items

arthurdm:
These items were missing from the OpenAPI Spec POM when compared with the REST Client Spec POM.


leochr:
@arthurdm looks good. thanks.

PR #156, opened 2017-12-13
Oas config tests

janamanoharan:


tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigServersTest.java
```diff
@@ -0,0 +1,54 @@
...
+import io.restassured.response.ValidatableResponse;
+
+public class OASConfigServersTest extends AppTestBase {
+    private ValidatableResponse vr;
+    
+    @Deployment(name = "airlines")
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, "airlines.war")
+                .addPackages(true, "org.eclipse.microprofile.openapi.apps.airlines")
+                .addAsManifestResource("package-microprofile-config.properties", "microprofile-config.properties");
```
arthurdm: did you mean to use `server-microprofile-config.properties` instead?

janamanoharan: yes, thanks for the catch

PR #171, opened 2017-12-14
Remove schema composition tests.

turkeylurkey:
Signed-off-by: Paul Gooderham <turkeyonmarblerye@gmail.com>


Azquelt:
Does anyone remember why these tests were removed?

I was about to add similar tests for the new properties we're adding to support OpenAPI 3.1

PR #189, opened 2018-01-25
Improved the assertions in the disable-scanning tck test.

### LINKED ISSUE #187: Liberty-specific assertion in OASConfigScanDisableTest
[The following code](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigScanDisableTest.java#L43-L50) can be found in the **OASConfigScanDisableTest** test:

```java
    @RunAsClient
    @Test(dataProvider = "formatProvider")
    public void testScanDisable(String type) throws InterruptedException {
        vr = callEndpoint(type);
        vr.body("openapi", equalTo("3.0.0"));
        vr.body("info.title", equalTo("Liberty APIs"));
        vr.body("info.version", equalTo("1.0"));
        vr.body("paths", aMapWithSize(0));
        
    }
```

I suggest only asserting on the openapi version (and even then looking for 3.0.x rather than 3.0.0) and the paths.  Perhaps also check for empty/missing components, servers, tags, etc?

EricWittmann:
Fixes #187

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigScanDisableTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigScanDisableTest.java
@@ -18,6 +18,8 @@ package org.eclipse.microprofile.openapi.tck;
 
 
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.collection.IsMapWithSize.aMapWithSize;
 
 import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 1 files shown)


arturdzm:
I was thinking adding default info needs to be set by the reader and shouldn't be null. This is so we can support pure JAX-RS apps without @Info annotation and still produce valid document.
I was thinking using something like this:
title: Deployed  API's
version: 1.0.0

EricWittmann:
My mistake - I didn't realize that `info` was a required property (along with `info.title` and `info.version`).  I will update the PR accordingly!

EricWittmann:
OK @arturdzm - I've pushed some changes that hopefully make more sense.  :)  Thanks!

arthurdm:
hey @EricWittmann - thanks for the PR.   

I would suggest that we don't check for servers being null, as that can leave it up to implementors if they want to have a default servers - for example, a server with URL that matches where the app is being exposed from (WebSphere Liberty does this).

If you can also rebase your changes to fix the conflicts that would be awesome.

EricWittmann:
Closing this PR - I submitted a new one to replace it.

PR #190, opened 2018-01-26
Updated all @ExampleObjects to have a name property

### LINKED ISSUE #184: Invalid assertions in AirlinesAppTest.testExampleObject()
In the `AirlinesAppTest` there is a method called `testExampleObject` with [the following two assertions](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java#L658-L662):

```java
// Example in Content
vr.body("paths.'/reviews/{user}'.parameters.find{ it.name=='user'}.content.examples.value", hasItem("bsmith"));

// Example in Parameter
vr.body("paths.'/reviews/{user}'.parameters.find{ it.name=='user'}.examples.value", hasItems("bsmith", "pat@example.com"));
```

I don't think either of these paths are possible in OAI 3.  For one thing, a `Content` object in OAI is always a map of `MediaType` objects.  As a result, the content property in the first assertion would not have an `examples` property.  It should instead be a map of media types.  The examples can then be found in the media type object.

I also don't think that `value` is a valid property for examples in this context.  It suggest that `value` is a property of the `examples` object.  But `examples` I think is a map of `Example` objects, so it seems there should be something else in there.

I've run out of steam today - I'll try to come up with what I think the resulting OAI JSON (or yaml) should be given the test classes.  I believe the assertion is trying to test [the following test app code](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java#L221-L229):

```java
@Parameter(
      name = "user",
      description = "username of the user for the reviews",
      required = true,
      in = ParameterIn.PATH,
      content = @Content(
          examples = @ExampleObject(
              value = "bsmith")),
      examples = { @ExampleObject("bsmith"), @ExampleObject("pat@example.com")})
```

EricWittmann:
Updated AirlinesAppTest.testExampleObject to fix a few assertion paths.

Fixes #184 

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
@@ -189,6 +189,7 @@ public class ReviewResource {
                 in = ParameterIn.PATH,
                 content = @Content(
                     examples = @ExampleObject(
+                        name = "example",
                         value = "1")))
             @PathParam("id") int id){
         Review review = reviews.get(id);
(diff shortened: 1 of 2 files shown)


EricWittmann:
Closing this PR, replaced by https://github.com/eclipse/microprofile-open-api/pull/193

Issue #198, opened 2018-01-31
@Schema.implementation -> $ref should be vendor specific


EricWittmann:
Several TCK tests assume that the following annotation will result in a $ref in the OAI document:

```
@Schema(
    implementation=BeanClass.class
)
```

For a variety of reasons discussed on yesterday's call, we have decided that the TCK should not require vendors to create a reference in the above case.  It's acceptable for a vendor to instead inline the schema definition.

For this reason, we should change the TCK apps to use `ref` instead of `implementation` if we really want to test references.

arthurdm:
Closed by PR #199

EricWittmann:
PR 199 closed issue 198!  SOOOO close!

PR #199, opened 2018-01-31
Fixed various problematic @Schema reference assertions.

EricWittmann:
Hey everyone.  Based on our conversations yesterday here is a proposal for fixes to all the $ref schemas throughout the TCK.  In some cases I had to fix the assertions (separate conversation about why) while in others I changed the annotation (from `implementation` to `ref`).  

Have a look and offer opinions here.  :)

I'll open a separate issue to discuss why I had to alter some of the assertions rather than the `@Schema` annotations.


EricWittmann:
Note:  this fixes issue #198

Issue #197, opened 2018-01-30
TCK test ModelReaderAppTest has an unused method: testAvailabilityGetParamater


EricWittmann:

arthurdm:
@leochr

leochr:
Checked with Anna Safonov who authored the ModelReaderAppTest and she believes that the test is missing the annotation tags. It's meant to test the construction of Parameters via programming models. Anna will make the necessary changes, verify that the test is correct/passes and make a PR.

EricWittmann:
For reference:  https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppTest.java#L156-L183

arthurdm:
Fixed by PR #202

PR #202, opened 2018-01-31
Issue 197: Configured testAvailabilityGetParameter to run with the test suite

a-saf:
Signed-off-by: Anna Safonov <anna.safonov@ibm.com>

This PR is for issue #197

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppTest.java
@@ -153,7 +153,10 @@ public class ModelReaderAppTest extends AppTestBase {
         vr.body("paths.'/availability'.get.responses.'404'.description", equalTo("No available flights found"));
     }
 
-    private void testAvailabilityGetParamater(ValidatableResponse vr) {
+    @RunAsClient
+    @Test(dataProvider = "formatProvider")
+    public void testAvailabilityGetParameter(String type) {
+        ValidatableResponse vr = callEndpoint(type);
         String availabilityParameters = "paths.'/availability'.get.parameters";
 
         vr.body(availabilityParameters, hasSize(6));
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppTest.java
```diff
@@ -153,15 +153,19 @@ public void testAPIResponse(String type) {
...
+    @Test(dataProvider = "formatProvider")
+    public void testAvailabilityGetParameter(String type) {
+        ValidatableResponse vr = callEndpoint(type);
         String availabilityParameters = "paths.'/availability'.get.parameters";
 
         vr.body(availabilityParameters, hasSize(6));
         vr.body(availabilityParameters + ".findAll { it }.name",
                 hasItems("departureDate", "airportFrom", "returningDate", "airportTo", "numberOfAdults", "numberOfChildren"));
+        
+        vr.body(availabilityParameters + ".findAll { it.ref == '#/components/parameters/departureDate'}", notNullValue());
```
leochr: I believe it should be 'it.$ref' instead of 'it.ref'.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppTest.java
```diff
@@ -153,15 +153,19 @@ public void testAPIResponse(String type) {
         vr.body("paths.'/availability'.get.responses.'404'.description", equalTo("No available flights found"));
     }
 
-    private void testAvailabilityGetParamater(ValidatableResponse vr) {
+    @RunAsClient
+    @Test(dataProvider = "formatProvider")
+    public void testAvailabilityGetParameter(String type) {
+        ValidatableResponse vr = callEndpoint(type);
         String availabilityParameters = "paths.'/availability'.get.parameters";
```
leochr: This endpoint should be changed to /availabilityModel (to match the change to the app).

The size check (below) will fail when you make that change because departureDate is a reference and wouldn't have a 'name'. So adjust the size check as well as remove departureDate from the 'hasItems' check.

leochr: Actually the size check wouldn't fail (it would still be 6 parameters). So only the hasItems check would have to be altered.

a-saf: The size of the collection will still be 6, but the check for the name will fail with departureDate there, so I'll remove that.

arthurdm:
Thanks @a-saf

PR #204, opened 2018-02-05
using MavenResolverSystem to resolve dependencies

janamanoharan:


tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
```diff
@@ -29,6 +29,8 @@
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.testng.Arquillian;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.resolver.api.maven.MavenResolverSystem; //maven.MavenDependencyResolver;
```
arthurdm: can we remove that comment at the end of the line?

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
```diff
@@ -48,20 +50,24 @@
...
     private static Map<String, String> headers = new HashMap<>();
-
+        
     @Deployment(name = "proxy")
     public static WebArchive createProxy() {
+        
+        MavenResolverSystem resolver = Maven.resolver();  
+        resolver.loadPomFromFile("pom.xml");  
+        File[] files = resolver.resolve("org.apache.httpcomponents:httpcore:4.4.4",
+                "com.fasterxml.jackson.core:jackson-core:2.8.6",
```
arthurdm: In the `tck/pom` we have variables such as {{httpclient.version} and {jackson.version}, but here we're hardcoding them.  Can we investigate a way to use the variables here too?  That way we only need to change a single place when updating versions in the future.

arthurdm:
Hey @janamanoharan - thanks for the PR.  I am closing this one though, as PR #206 modified the TCK so that we don't need the proxy.war anymore.

PR #205, opened 2018-02-05
Add explicit media type to a callback

arturdzm:
Since Callback responses and request body can have different media type from the method it applies to. Its up to the vendor if the method's or class' @Produces applies in this case so its best to explicitly state mediaType


EricWittmann:
+1 makes sense.

arthurdm:
I think that's a reasonable update to the TCK, however, in other places we have assertions that a lack of media type in a `Content` annotation results in `*/*` being assigned as a media type.   So I just want to make sure that we aren't inconsistently applying defaults to instances of `Content`.

arturdzm:
So in this case default value produced was */* so the test was failing, in cases where media type is missing for request body and responses we try to grab one from method or class Produces and Consumes. However since this a callback we decided not to do that. In the case of this test it means that server expects response of callback to be in certain format. This format is not necessary will match current method content type.

Issue #217, opened 2018-05-17
TCK Bugs?


MattGill98:
Hi, I was hoping someone wouldn't mind helping me clarify a few problems I've encountered when working with the TCK.

Firstly, the adding of the `microprofile-config.properties` seems strange. In the Config API TCK, this file is added using this line:
```
.addAsWebInfResource(mpConfig, "classes/META-INF/microprofile-config.properties");
```
This adds the file to `WEB-INF/classes/META-INF`, which is on the application class path.
In the OpenAPI TCK however, this file is added using this line:
```
.addAsManifestResource("microprofile-config.properties", "microprofile-config.properties");
```
This adds the file to `META-INF`, which isn't on the application class path. Since the Config API doesn't test for this, it's behaviour is presumably undefined. Should this be an added test to the Config API TCK, or a required change in the OpenAPI TCK?

My second problem is regarding the AirlinesAppTest. This application seems to be invalid, and as such may not deploy to all application servers. The endpoint in question involves these two methods in the `ReviewResource`:
```
@GET
@Path("{user}")
@Produces("application/json")
public Response getReviewByUser(@PathParam("user") String user) {}
```
```
@GET
@Path("{id}")
@Produces("application/json")
public Response getReviewById(@PathParam("id") String id) {}
```
There's no problem with this from the perspective of the OpenAPI, as they will appear as different endpoints. For the JAX-RS server though these endpoints are functionally the same, and so may not deploy correctly on all application servers. In my opinion this should not cause a failed test however, and maybe one of these endpoints should be fixed to have a different `@Produces` type. Is this an accurate evaluation?

Thanks in advance,

Matt

arthurdm:
hey @MattGill98 - thanks a lot for the feedback.

For the first issue, I believe it's a gap in the MP Config TCK, as [its design](https://github.com/eclipse/microprofile-config#design) outlines that `META-INF/microprofile-config.properties` should be picked up by the config processor.

I agree with your observation of the second issue.  I propose that we update the path of `getReviewByUser` to be ` @Path("users/{user}")`, which would solve this.  

Thoughts?

MattGill98:
Hi Arthur,

Thank you for your prompt response. That's not a problem. That being the case, I'll raise an issue for the Config API project!

That would work absolutely fine. I'll make a pull request making this change.

Kind regards,

Matt

MattGill98:
Hi @arthurdm,

[...]

PR #218, opened 2018-05-17
Fixed Endpoint Naming Clash

MattGill98:
PR to address issue: https://github.com/eclipse/microprofile-open-api/issues/217.

This change fixes an endpoint path so that the application contains no identical endpoints.


MattGill98:
Thanks!

PR #220, opened 2018-05-18
Moved META-INF resources into WEB-INF/classes/META-INF

MattGill98:
PR to address issue: #217.

This change moves resources from `META-INF`, which is not explicitly defined to be correct in the Config API spec, to `WEB-INF/classes/META-INF`, which [is](https://github.com/eclipse/microprofile-config/issues/268).

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigWebInf.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigWebInf.java
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


MattGill98:
That makes sense. Would you rather I made your requested changes now to this PR, or waited for the result of #221?

arthurdm:
@MattGill98 -maybe it's good to wait until our next hangout (May 28th), where we should resolve #221

arthurdm:
btw - that meeting now moved to June 5th.

arthurdm:
hey @MattGill98 - thanks for your patience on this PR.  Now that we have resolved #221, my review comment is that you can certainly add more tests to the TCK that ensure `microprofile-config.properties` can also be set inside `WEB-INF`, but those should be additional to the current ones (instead of the replacing them).  

Also, please note that `openapi` static docs must remain in `META-INF`, as per the latest clarification in the spec.

MattGill98:
Hi @arthurdm,

I apologise for the late turnaround. Thank you for your clarification, I've added a new test that runs the `OASConfigScanClassTest`, but with a new deployment placing only the `microprofile-config.properties` into the `WEB-INF/classes/META-INF` directory. I hope this meets the requirements more accurately.

arthurdm:
thanks @MattGill98 - the class `OASConfigWebInf` looks good.  Was there a corresponding test that didn't get committed? I only see 1 class in the `files changed` tab.

MattGill98:
My bad, I'd made a few changes at once and forgot to change the file name! The idea is that it's a new test that just runs the same tests in `OASConfigScanClassTest`, but instead puts the microprofile-config file into the `WEB-INF/classes/META-INF` directory.

Issue #234, opened 2018-07-03
Questions about org.eclipse.microprofile.openapi.models.PathItem


jmini:
I am new to the Microprofile project, I hope this is the appropriate channel to start a discussion.

How definitive are the model interfaces introduced with version MicroProfile release 3.1?
I understand that Swagger Core was used as reference, but maybe some stuff can be improved when establishing a standard.

Today I have worked a lot with `PathItem` and `Operation` (I work with swagger-core) and I was curious to see how you guys did solve this.

Here 2 questions about `org.eclipse.microprofile.openapi.models.PathItem`.

Thank you in advance for the discussion.

---

**Question 1:** why are the convenience methods that provides all operations as list or as map called `readOperations()` and `readOperationsMap()`:

https://github.com/eclipse/microprofile-open-api/blob/bffd7ace546bd964cb46d29dcf2114c8ac28fe80/api/src/main/java/org/eclipse/microprofile/openapi/models/PathItem.java#L266-L278

I like the way the Kaizen-Parser solves this by providing a map (they do not have the enum for HTTP operation type):

```java
	Map<String, Operation> getOperations();
```
([source](https://github.com/RepreZen/KaiZen-OpenApi-Parser/blob/ad8fa48166e0540ba09ed028292a3a06aa512516/kaizen-openapi-parser/src/main/java/com/reprezen/kaizen/oasparser/model3/Path.java#L31))

If we take the best of both words:

```java
Map<PathItem.HttpMethod, Operation> getOperations();
```

With the constraint that the returned `Map` should never be `null` and that the map is a read-only access to the operations. (could be indicated in the Javadoc)

If you just want the collection of operations `getOperations().values()` can be called. I do not think that a method returning `List<Operation>` is requested.

---

**Question 2:** In Swagger-Core the `String` path of an `PathItem` is an important information that belongs in my opinion to the `PathItem` object, but that is missing the `Operation` object.
 
To access to this information you need the corresponding `Entry<String, PathItem>` of the `Paths` map holding the `PathItem`.

Because this information is not present, you need always to pass it with in each `PathItem` instance in of your methods:

```java
void printOperationsInfo(String path, PathItem path);
```

Or to have your own holder class: 

```java
static class PathItemWithPath {
    final String pathString;
    final PathItem path;

    public PathItemWithPath(final String pathString, final PathItem path) {
        super();
        this.pathString = pathString;
        this.path = path;
    }
```

The Kaizen-Parser Model did solve this with:

[...]

PR #235, opened 2018-07-12
Change methods to access Operations in PathItems

jmini:
See original discussion in issue #234 (point `1.`)

I was surprised to not find any tests for the model in the TCK.

I understand that it focus on App-Server stuff (equivalent of ` 	swagger-jaxrs2`) but you have a set of constraints on the model interfaces as well.

For example, With:

```
PathItem p;
Operation o;
```

You can do:
```java
p.readOperationsMap().put(HttpMethod.POST, o);
```
Or:
```
p.setPOST(o);
```

Does the API tells if they are equivalent or not?

If you take the swagger-core implementation it is not. 
`readOperationsMap()` is somehow a view map (it should be unmodifiable in my opinion, this is an other story).

On the other side, doing this on the Kaizen model seems to update the the POST operation.

I think that going over the setter should be the recommended way.
Forcing the `getOperations()` to be unmodifiable is a to restrictive, but allowing modification is also a problem for implementation that do not keep the information inside a Map.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/models/PathItem.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/models/PathItem.java
@@ -266,16 +266,27 @@ public interface PathItem extends Constructible, Extensible, Reference<PathItem>
     /**
      * Returns a list of all the operations for this path item.
      * 
+     * @deprecated since 1.1, use @link {@link Map#values()} on {@link #getOperations()} instead
      * @return a list of all the operations for this path item
      **/
+    @Deprecated
     List<Operation> readOperations();
 
     /**
      * Returns a map with all the operations for this path where the keys are HttpMethods.
      * 
+     * @deprecated since 1.1, use {@link #getOperations()} instead
      * @return a map with all the operations for this path where the keys are HttpMethods
      **/
+    @Deprecated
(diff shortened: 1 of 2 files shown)


arthurdm:
thanks @jmini.  Have you looked at this TCK test? https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java   It tests the methods of the model classes.

I agree about the lack of description on how these methods interact.  I think that making the methods that return a map as immutable is a good idea.  I will open an issue for it.

jmini:
@arthurdm: I do not see any `readOperationsMap()` call in `ModelConstructionTest.java`.

jmini:
I have added some tests to `pathItemTest()`. Please review them carefully, because it is the first time I write some TCK tests.

2 questions:

* Should the deprecated methods provide a default implementation that delegates to `getOperations()`?
* Should my TCK tests also test the deprecated method?

arthurdm:
I don't think it's necessary to test or add default impls for the deprecated methods.

jmini:
I think I have updated the test as you have suggested.

arthurdm:
looks great, thanks @jmini!

jmini:
I had an additional idea: why not provide a default implementation for the 2 olds methods? See #243.

Issue #216, opened 2018-05-15
Support singular example in Content annotation


arthurdm:
At the moment our `@Content` annotation only supports an array of `ExampleObject`, however, our `Content` **model** correctly supports both the array and the single `example`, as per the OAS3 spec.  

This workitem will include the single example inside the `@Content` annotation, so that a simple / single example can be given without affecting the underlying `@Schema`, since it could be referenced from different places.

We should add a corresponding TCK test too.

EricWittmann:
I can take this one.

ivanjunckes:
Hi Arthur, I was taking a look at this today and I probably didn't quite understand the description.

These are the references to the example object I found. All of them are using arrays and they support one value as well.
![image](https://user-images.githubusercontent.com/2260551/41824639-eb20ff3a-77e9-11e8-8c47-6b42f660eb47.png)

Not sure I get when you say it only supports array. I am probably missing something, could you please point to the code so I can better understand?

Thanks.

arthurdm:
Hey @ivanjunckes - Currently the `@Content` annotation just support this: https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Content.java#L47

So we need to add an additional method with the signature `String example();`, such as done in here: https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java#L265

arthurdm:
Closing - finished in PR #239

PR #237, opened 2018-07-13
#216 - Fix content to have example

ivanjunckes:


EricWittmann:
OK - a couple of things here.  First, in order to contribute to **eclipse/microprofile-open-api** you must use the git "Signed Off By" feature for your commits.  See here:

https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=5b47ffc235a1f

So unfortunately I think you'll need to re-do the commit and resubmit.

That might be OK though, because the spec says that "example" and "examples" are mutually exclusive properties.  So I'm wondering if the TCK test class `ReviewResource` should have a `@Content` annotation with **both** `example` and `examples` properties.  It's unclear to me whether the TCK should be requiring implementations process annotations in that way, or if implementations are free to process the annotations in a way that ensures a valid OpenAPI document result.

In other words, an implementation might want to ignore the `example` property in cases where an `examples` property is also provided.

@arthurdm thoughts?

I looked at the other tests for examples here:  https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java#L650-L665

And I don't see any other tests at all for the `example` property.

ivanjunckes:
Opened another one, please see if it is ok with the IP.

https://github.com/eclipse/microprofile-open-api/pull/239

EricWittmann:
I think this one can be closed now - the new PR is great.

PR #239, opened 2018-07-13
#216 - Add example to content

ivanjunckes:
Signed-off-by: ivanjunckes <ijunckes@tomitribe.com>

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Content.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Content.java
@@ -46,6 +46,14 @@ public @interface Content {
      **/
     ExampleObject[] examples() default {};
 
+
+    /**
+     * A free-form property to include an example of an instance for this content.
+     *
+     * @return an example of this content
+     **/
+    String example() default "";
+
     /**
      * The schema defining the type used for the request body.
      *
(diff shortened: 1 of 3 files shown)


EricWittmann:
Great, thanks for opening a new PR, Ivan!  So now the only question I have is whether the TCK should have a separate test for `example` or whether sharing it with `examples` is OK.

What do you think, @arthurdm ?

arthurdm:
thanks @ivanjunckes !     I agree with @EricWittmann - maybe you can simply change the value of the singular example to something like `jsmith`?   Just so that we can ensure the singular and plural examples aren't overriding each other.

EricWittmann:
I actually meant that I think we need a TCK test where `@Content` doesn't contain **both** the `example` and `examples` properties.  Since it's an OAI spec violation to use both properties in the same `Content` element.

My suggestion is to either add another method to the JAX-RS resource, or find an `@Content` annotation that doesn't have an example defined.

arthurdm:
good point @EricWittmann - I had forgotten about that restriction.

ivanjunckes:
Cool I will fix that.

ivanjunckes:
Fixed, I removed one of the "examples" that were not being used anywhere in the TCK, and used it to test "example". I hope this is ok.

arthurdm:
thanks @ivanjunckes!

Issue #236, opened 2018-07-12
Models: using default methods for builder methods?


jmini:
In the `org.eclipse.microprofile.openapi.models` package and sub-packages, the builder-pattern is used:

```java
    T getXyzAbcd();

    void setXyzAbcd(T xyzAbcd);

    U ayzAbcd(T xyzAbcd);
```

The `ayzAbcd(T)` method is just a builder method.

Could this method be a default method?

```java
    default U XyzAbcd(T xyzAbcd) {
        setXyzAbcd(xyzAbcd);
        return this;
    }
```

In my opinion it would ease the work of people who wants to create an implementation of the model interfaces.

Or is there a reason to not do this?

arthurdm:
We agreed with this in today's hangout.  This issue will be used to drive the proposed work of adding the default impls for these simple builder methods.

jmini:
I have created an example in https://github.com/eclipse/microprofile-open-api/pull/242 for the `Operation` class. Before I extend this pattern to all builders methods, I want to be sure that this is the idea we have discussed here.

jmini:
> This issue will be used to drive the proposed work of adding the default impls for these simple builder methods.

Have you other changes than the one in #242 in mind?

arthurdm:
that was all.  thanks again @jmini!

PR #242, opened 2018-07-17
Default implementation for builder methods

jmini:
See Issue #236

Signed-off-by: Jeremie Bresson <dev@jmini.fr>


jmini:
This is an example for the `Operation` class. Please leave a comment before I start to apply this pattern to all builders methods. I need to be sure that this change corresponds to the discussion.

arthurdm:
looks good @jmini - thx for working on this.

jmini:
I just updated the PR to handle all cases.

api/src/main/java/org/eclipse/microprofile/openapi/models/Reference.java
```diff
@@ -64,6 +64,9 @@
      * @param ref a reference to a T object in the components in this OpenAPI document
      * @return the current instance
      **/
-    T ref(String ref);
+    default T ref(String ref) {
+        setRef(ref);
+        return (T) this;
```
jmini: Here I am not sure about this, because it generate this warning:

> Type safety: Unchecked cast from Reference<T> to T

Should I add a `@SuppressWarnings("unchecked")` to the method?

An alternative is to do it like that:

```java
    default T ref(String ref) {
        setRef(ref);
        @SuppressWarnings("unchecked")
        T t = (T) this;
        return t;
    }
```

jmini:
I have update the PR.

> the `Reference` interface is an interesting one, because we don't necessarily need a `ReferenceImpl` - instead these methods are implemented in the models that are reference-able, like the PathItem..so in that case it doesn't need generics. Here's how Open Liberty does it: https://github.com/OpenLiberty/open-liberty/blob/c2c03fc012c4750f787fd0402385d3ab95984b14/dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/impl/model/PathItemImpl.java#L349-L353

With the proposed implementation:

```java
    default T ref(String ref) {
        setRef(ref);
        @SuppressWarnings("unchecked")
        T t = (T) this;
        return t;
    }
```
Having to cast the return value to `(T)` is just to make the java compiler happy.
The default method implementation works exactly as expected.

With an implementation like this:
```java
public class SchemaImpl implements Schema {
    //...
}
```
Which does not redefine the default methods proposed in this PR.

You can then write:
```java
Schema schema = new SchemaImpl()
		.ref("bla")
		.description("blabla");
```
(Which makes no sense from an OpenAPI point of view, but this is an other story).
Everything works as expected.

PR #250, opened 2018-07-31
Change Extensible#addExtension(String, Object) to builder method

### LINKED ISSUE #246: Models: Extensible.addExtension(..) should be a builder method
For consistency with other patterns:

All methods that allow to add an item in a list or in a collection are builder methods.

It should be also the case for:
```
org.eclipse.microprofile.openapi.models.Extensible.addExtension(String, Object)
```

jmini:
Fixes #246


jmini:
This PR creates a compile error in the TCK project. It is fixed with #252.

jmini:
Follow up PR for the TCK (add test and fix warnings): #276

Issue #248, opened 2018-07-25
Models: do we need the interfaces extending Map?


jmini:
There are several model interfaces:
`Paths`, `ServerVariables`, `Content`, `APIResponses`, `Scopes`, `Callback`, 
that extends, `Map<String, T>`

My points are:

1/ Those Interfaces are really strange, when working with it, you never can not use stuff like:

* `Arrays.asList(..)`
* Streams + Collectors

2/ To implement #240 we might need to remove those interface. How do you create an immutable version of `Paths`

3/ The KaiZen-Parser models do not use such a construct.

jmini:
@EricWittmann : thank you for the "extensible" pointer.

I divided my list into two categories:

**1) Not extensible.**

No extends Extensible in the current code:
```
org.eclipse.microprofile.openapi.models.media.Content
org.eclipse.microprofile.openapi.models.responses.APIResponses
```

For:
```
org.eclipse.microprofile.openapi.models.servers.ServerVariables
```
As discussed in  #245 it is marked as extensible for the moment, but this is wrong according to the spec.


**2) Extensible.**

```
org.eclipse.microprofile.openapi.models.Paths
org.eclipse.microprofile.openapi.models.callbacks.Callback
org.eclipse.microprofile.openapi.models.security.Scopes
```

Extensible in the current code and marked as such in the OpenAPI-Spec.

I would like to understand how an extension on the `Paths` or `Scopes` looks like. A comparison with KaiZen is also requested.

arthurdm:
+1 for replacing the non-extensible model interfaces with a map version in the appropriate encapsulating model.

For the extensible (as per OAS3 spec) model interfaces, I propose that we don't extend from `Map`, and instead just contain a map as an internal field - exposing the non-mutable collection, and appropriate `add` and `remove` methods.

arthurdm:
For `1.1` we can add the `map` fields, `add` and `remove` methods and put a deprecating notice at the appropriate places.  

I opened https://github.com/eclipse/microprofile-open-api/issues/251 to start tracking future 2.0 candidates - I think this one fits that.

jmini:
`org.eclipse.microprofile.openapi.models.responses.APIResponses` are extensible, see #254

jmini:
> For 1.1 we can add the map fields, add and remove methods and put a deprecating notice at the appropriate places.

I am OK with this, can we sketch how the methods should look like?

Example with `Paths`:

Current situation:
[...]

PR #266, opened 2018-09-03
Add manipulations methods to Paths

jmini:
Implementation of #248 for `Paths` for the `1.1` version to start a discussion.

TCK coverage could be improved.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
@@ -17,6 +17,7 @@
 package org.eclipse.microprofile.openapi.tck;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNotSame;
 import static org.testng.Assert.assertNull;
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
```diff
@@ -311,12 +311,16 @@ public void pathsTest() {
         final String pathItemKey = "myPathItem";
```
jmini: Not directly related to this change.

Current javadoc of `addPathItem(..)` says:
> name a path name in the format valid for a Paths object. The field name MUST begin with a slash.

But the TCK adds a `PathItem` without leading slash.

Should `pathItemKey` be changed to add `"/myPathItem"` instead?

---

Should an other test be created that ensure that `"somePath"` can not be added? In this case, what is the type of exception that should be expected by the TCK test?

---
Current Javadoc (not changed with this PR) for reference:
https://github.com/eclipse/microprofile-open-api/blob/8e0f2479a00470911d64ae13a59c5cd78f4240c9/api/src/main/java/org/eclipse/microprofile/openapi/models/Paths.java#L34-L41

jmini: I have changed my mind on this matter:

The models can contain malformed OpenAPI instances. 

If at TCK level we enforce the OpenAPI Spec to be valid, this means that the models classes can no hold malformed Spec. This is used by Parsers to report "Errors" and "Warnings".

arthurdm: good point @jmini.   My suggestion is to make the TCK test a valid OAS instance (so using `/myPathItem`).   We haven't covered in the MP spec how to deal with errors and warnings so at the moment it's implementation dependent.  This is probably a great topic for discussion though, and potential `2.0` item.

arthurdm:
We'll try to come to consensus on this via https://github.com/eclipse/microprofile-open-api/issues/248 in the next hangout.

PR #249, opened 2018-07-30
Add missing methods: addHeader(..) and addCallback(..)

jmini:
As discussed in Issue #229 this PR adds:

```
org.eclipse.microprofile.openapi.models.media.Encoding.addHeader(String, Header)
org.eclipse.microprofile.openapi.models.Operation.addCallback(Callback)
```

PR #272, opened 2018-09-18
Add missing key parameter in Operation#addCallback()

jmini:
In PR https://github.com/eclipse/microprofile-open-api/pull/249 I have introduced:

```
org.eclipse.microprofile.openapi.models.Operation.addCallback(Callback)
```

But `callbacks` in `Operation` is a `Map<String, Callback>`, so the method should be:

```
org.eclipse.microprofile.openapi.models.Operation.addCallback(String, Callback)
```

This PR adds the missing key parameter.

---

`1.1-SNAPSHOT` was not published yet, so I hope it is OK to fix this now.

---

cc: @EricWittmann, @arthurdm


jmini:
Thank you for the review.

I have modified Javadoc according to comment and rebased the PR on top of master.

Issue #270, opened 2018-09-16
Running TCK `ModelConstructionTest` as standalone test (without having a deployable)?


jmini:
I have created an implementation for the models interface (and only for those). I am not interested by the "Application server" part of the project for my implementation.

I wanted to validate my project against the relevant part of the TCK (I identified at least `ModelConstructionTest` for that matter).

I have created a TestNG test suite:

```xml
<?xml version = "1.0" encoding = "UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >

<suite name = "Suite">

   <test name = "tcktests">
      <classes>
         <class name = "org.eclipse.microprofile.openapi.tck.ModelConstructionTest" />
      </classes>
   </test>

</suite>
```

When I try to run it, I get this exception:

```
org.jboss.arquillian.container.test.impl.client.deployment.ValidationException: DeploymentScenario contains a target (_DEFAULT_) not matching any defined Container in the registry.
Please include at least 1 Deployable Container on your Classpath.
	at org.jboss.arquillian.container.test.impl.client.deployment.DeploymentGenerator.throwNoContainerFound(DeploymentGenerator.java:250)
	at org.jboss.arquillian.container.test.impl.client.deployment.DeploymentGenerator.throwTargetNotFoundValidationException(DeploymentGenerator.java:243)
	at org.jboss.arquillian.container.test.impl.client.deployment.DeploymentGenerator.validate(DeploymentGenerator.java:102)
	at org.jboss.arquillian.container.test.impl.client.deployment.DeploymentGenerator.generateDeployment(DeploymentGenerator.java:84)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at ...
```

I have the feeling that `ModelConstructionTest` should not extends `Arquillian`, because it is not necessary.

Can you please confirm, comment, provide feedback?

jmini:
This is the commit I have in mind: https://github.com/jmini/microprofile-open-api/commit/05052b3d2dc9b5ef6b9e02ba6e642e4066f6a57a

I wait for feedback to open the PR.

jmini:
I have opened https://github.com/eclipse/microprofile-open-api/pull/274 to discuss it in a PR.

arthurdm:
Fixed by #274

PR #274, opened 2018-09-24
Remove Arquillian feature from ModelConstructionTest

jmini:
See #270.

This is requested for my use case (I am just implementing the model part of the spec), I can not imagine it breaks other implementations (the one that really have deployments to test), but I am not 100% sure about it.


tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
```diff
@@ -71,18 +71,14 @@
...
-import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.annotations.Test;
 
 /**
  * This test covers construction of the OpenAPI model. It verifies that the implementation can
  * create instances of all of the Constructible interfaces and then invokes methods (including
  * getters, setters and builders) on those instances to verify that they behave correctly.
  */
-public class ModelConstructionTest extends Arquillian {
```
jmini: If it helps other use cases, this line does not need to be modified.

Issue #257, opened 2018-08-20
Schema.getAdditionalProperties(): return something better as Object?


jmini:
Per spec: `additionalProperties` in a Schema object can be an Object or a Boolean.
See https://github.com/OAI/OpenAPI-Specification/issues/668#issuecomment-281837896

In the code, `Object` is returned:
https://github.com/eclipse/microprofile-open-api/blob/c2a26360d460bfe07b30c83ba4fe6f9cc73543c6/api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java#L641

This is not really user-friendly because the user needs to know that this `Object` can be a `Schema` or a `Boolean` and write code with `instanceOf` to reflect this.

---

Kaizen Model has 2 getters for this:

* `Boolean getAdditionalProperties()`
* `Schema getAdditionalPropertiesSchema()`

I do not like this either, but it is better in a sense that no `instanceOf` is needed.

[kaizen-parser Schema.java](https://github.com/RepreZen/KaiZen-OpenApi-Parser/blob/9295db46b47406aaa7046ceca5ef03d331cd6a51/kaizen-openapi-parser/src/main/java/com/reprezen/kaizen/oasparser/model3/Schema.java#L314-L332)

---

What is the typical java way of telling `Schama` or `Boolean` (a sort of Union Types)?
A new Interface `AdditionalProperties` with 2 getters returning an optional: 

* `Optional<Schema> getAsSchema()` 
* `Optional<Boolean> getAsBoolean()`

jmini:
Thank you for discussing this in the Hangout, and thank you for reporting here the decision.

Current (1.0):
```
Object getAdditionalProperties()
```

---

Is the idea to have:

```
Boolean getAdditionalPropertiesBoolean()
Schema getAdditionalPropertiesSchema()
```

Then I understand that `getAdditionalProperties()` needs to be removed with 2.0 (and can be deprecated with 1.1).

I prefer this solution, because if users are using `Object o = s.getAdditionalProperties()` it is not clear that the method now only returns Boolean.

---

But you have mentioned:

```
Boolean getAdditionalProperties()
Schema getAdditionalPropertiesSchema()
```

Then the only modification is that `getAdditionalProperties()` was returning an `Object` with 1.0 and will return a `Boolean` with 2.0.

jmini:
Additional proposition:

We have also already 2 setters:
[...]

PR #277, opened 2018-09-25
Distinguish Schema and Boolean case for AdditionalProperties in Schema

jmini:
See Issue #257.

I think the TCK should test getter/setter/builder for additionalProperties in `ModelConstructionTest.schemaTest()`. I was surprised to see nothing there for the moment.
=> I will add a commit to this PR.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
@@ -474,6 +474,20 @@ public class ModelConstructionTest {
     public void schemaTest() {
         final Schema s = processConstructible(Schema.class);
         
+        final Schema ap = createConstructibleInstance(Schema.class);
+        checkSameObject(s, s.additionalPropertiesSchema(ap));
+        checkSameObject(ap, s.getAdditionalPropertiesSchema());
+        assertEquals(s.getAdditionalPropertiesBoolean(), null, "AdditionalProperties (Boolean type) is expected to be null");
+        checkSameObject(s, s.additionalPropertiesBoolean(Boolean.TRUE));
+        assertEquals(s.getAdditionalPropertiesBoolean(), Boolean.TRUE, "AdditionalProperties (Boolean type) is expected to be true");
+        assertEquals(s.getAdditionalPropertiesSchema(), null, "AdditionalProperties (Schema type) is expected to be null");
+        s.setAdditionalPropertiesBoolean(Boolean.FALSE);
+        assertEquals(s.getAdditionalPropertiesBoolean(), Boolean.FALSE, "AdditionalProperties (Boolean type) is expected to be false");
+        assertEquals(s.getAdditionalPropertiesSchema(), null, "AdditionalProperties (Schema type) is expected to be null");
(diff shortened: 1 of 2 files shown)


jmini:
TCK test added with 0bf0c2d

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
```diff
@@ -456,6 +456,20 @@ public void mediaTypeTest() {
     public void schemaTest() {
         final Schema s = processConstructible(Schema.class);
         
+        final Schema ap = createConstructibleInstance(Schema.class);
+        checkSameObject(s, s.additionalPropertiesSchema(ap));
+        checkSameObject(ap, s.getAdditionalPropertiesSchema());
+        assertEquals(s.getAdditionalPropertiesBoolean(), null, "AdditionalProperties (Boolean type) is expected to be null");
+        checkSameObject(s, s.additionalPropertiesBoolean(Boolean.TRUE));
+        assertEquals(s.getAdditionalPropertiesBoolean(), Boolean.TRUE, "AdditionalProperties (Boolean type) is expected to be true");
+        assertEquals(s.getAdditionalPropertiesSchema(), null, "AdditionalProperties (Schema type) is expected to be null");
```
arthurdm: I am trying to understand why the schema property would be `null` here, and not `ap`.  Is it because we set the boolean property, and therefore would have automatically set the schema property to null?  If so we should change the javadoc of the methods to specify this behaviour, and also change the default interface impl to set the other variant to null.

jmini:
This OpenAPI spec snippet shows the possible values for `additionalProperties`:
```yaml
  schemas:
    Test1:
      additionalProperties: false
    Test2:
      additionalProperties: true
    Test3:
      additionalProperties:
        type: object
        description: a person object
        properties:
          firstName:
            type: string
          lastName:
            type: string
```
(+ the null case, of course)

The `additionalProperties` is either a `Boolean` or a `Schema` but can not be both on the same time.

Because in Java we do not have union type, we decided to follow KaiZen approach: having getter,  setters and builder as if there was 2 members `additionalPropertiesBoolean` and `additionalPropertiesSchema`.

In comparison swagger model just use `Object` and the consumer needs to know that can be made with the value (call `instanceof Schema` or `instanceof Boolean` to see what the `additionalProperties` member really is).

In my opinion the implementation needs to decide how it wants to handle this case in the setter and getter for `additionalPropertiesBoolean`, `additionalPropertiesSchema`.

1/ They can decide to store it in an `Object` member, and then do an instanceof + cast in the getter.

2/ They can decide to have 2 members, but then they need to `null` the value in the setter when the other value is set.

3/ Or they use their own storage layer (I did not verify it, but my understanding is that Kaizen does store everything in the JSON data structure).

The default implementation method only does a “delegation to the setter”. Doing anything else in those case might break this rule. The implementation needs to do it right anyway, it is not the responsibility of the builder.

Does this make sense?

---

Maybe I need to write the TCK test in an other way.

---

I have tried to update Javadoc:

* I have added a precision in the getter.
* In the setter and builder there is already a note about the method being _“mutually exclusive”_ with the other variant. I have added a reference to the other method.

Issue #271, opened 2018-09-16
APIResponses defaultValue


jmini:
Some propositions for the `defaultValue` methods in `APIResponsesImpl`

Current:

```java
public APIResponse getDefault()
public void setDefaultValue(APIResponse defaultValue)
public APIResponses defaultValue(APIResponse defaultValue)
```

**First proposition:**

Rename the getter `getDefaultValue()` for consistency with the setter and the builder method.

**Second proposition:**

Specify at JavaDoc and TCK level that the methods for `DefaultValue` are only convenience methods that add a member to the `APIResponses` map with the `default` key (Constant `APIResponses.DEFAULT` already exists).

In other worlds:

```java
setDefaultValue(response);
```

Is the same as:
```java
addAPIResponse(APIResponses.DEFAULT, response);
```

With the same for the getter and the builder methods.
This should be covered at TCK level.

When we are at this, the keys in the TCK tests are strange:

https://github.com/eclipse/microprofile-open-api/blob/8e0f2479a00470911d64ae13a59c5cd78f4240c9/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java#L493-L508

From an OpenAPI-Spec point of view, valid key values are:

* HTTP codes
* HTTP codes with wildcards `2XX`, `3XX`
* `default`

See: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#patterned-fields-1

Suggested by @arthurdm:

> My suggestion is to make the TCK test a valid OAS instance

And I agree with this. Some implementation might be very strict about the accepted values.

arthurdm:
+1.  Discussed in the hangout on Sept 24.  Agree with changing TCK to use valid operation keys.  Also agreed to update the javadoc to make clear that `default` is a convenient key.     

For `1.1` add `getDefaultValue` and deprecate `getDefault`.  In `2.0` remove the deprecated method.

jmini:
PR for version `1.1`: #278. Feedback welcomed.

arthurdm:
`1.1` portion fixed by merged #278

arthurdm:
Awesome work @jmini - I think this one can be closed now?

jmini:
Yes everything is implemented. Thank you for review and merge.

PR #278, opened 2018-09-26
Define behavior of defaultValue in APIResponses

jmini:
See issue #271


jmini:
TCK Tests are pending.

jmini:
Commit rewritten to include TCK tests.

PR #283, opened 2018-10-03
Add manipulations methods for maps

jmini:
Add manipulations methods for maps in following interfaces:

* Paths
* Callback
* Content
* APIResponses
* Scopes
* SecurityRequirement
* ServerVariables

See Issue #248

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/models/Paths.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/models/Paths.java
@@ -41,14 +41,6 @@ public interface Paths extends Constructible, Extensible<Paths>, Map<String, Pat
     Paths addPathItem(String name, PathItem item);
 
     /**
-     * Adds all the path items to this Paths and return this instance of Paths
-     * 
-     * @param items a map containing the list of paths. Keys MUST begin with a slash.
-     * @return the current Paths instance
-     */
-    Paths addPathItems(Map<String, PathItem> items);
-
-    /**
      * Removes the given path item to this Paths.
      * 
      * @param name a path name that will be removed.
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/eclipse/microprofile/openapi/models/Paths.java
```diff
@@ -56,27 +48,38 @@
     void removePathItem(String name);
 
     /**
-     * Returns an immutable map of the path items.
+     * Returns a copy map (potentially immutable) of the path items.
```
arthurdm: minor thing, but in some of the other methods we have default implementations that rely on `getPathItems` to never return null.   Perhaps we should clarify in this javadoc that you must always return a non-null map, even if empty?

jmini:
@arthurdm: Your comment about the null safety is really interesting.

For the moment (with version `1.1`) having a `null` map is almost not possible because the container itself (`Paths`, `Callback`, `Content`, `APIResponses`, …) is the map.

My guess is that all implementations will do something like this:

```java
    public Map<String, PathItem> getPathItems() {
        return java.util.Collections.unmodifiableMap(this);
    }
```

But if we think on `2.0` this will be an other story.
For the moment there is no enforcement for any map getters to return empty map instead of `null`.
Example: `schema.getProperties()` can be null.

And from a Yaml point of view, this is also not the same.

Empty Map case:

```yaml
MyObject:
  type: object
  Description: Some generic Object
  properties: {}
```

Null case:

```yaml
MyObject:
  type: object
  Description: Some generic Object
```

At the end this express the same thing (no properties in this Schema), but I do not think that that the decision should be taken at API level. Each implementation can decide to if they prefer `null` or "empty map".

So I propose to change the convenience default implementation in the API.

What do you think?

PR #301, opened 2018-11-28
Adding SPI for platform APIs

### LINKED ISSUE #212: Investigate the addition of an SPI
The MP OpenAPI spec specifies how to fetch the OpenAPI v3 doc for a single application at `/openapi`.  However, the application server that runs this applications likely has other RESTful APIs that it wants to expose - not from applications, but from the server itself.  Examples include other MicroProfile specs, such as `/health` and `/metrics`, as well as other potential endpoints such as an OpenID Connect endpoint at `/oidc` to manage tokens.

This workitem explores the idea of an SPI that allows vendors to provide OpenAPI v3 definition for these extra "management" side APIs.  

Rough proposal:
(1)  a new Java interface (SPI), where 0..N providers inside the application server can implement:
```
interface OpenAPIProvider
   String getContextRoot()
   String getOpenAPI()
```

(2) a new endpoint, `openapi/providers` which returns a list of currently registered OpenAPIProvider's context roots.

(3) For each `{provider}` in (2), a new URL `openapi/providers/{provider}`.

-----  Example -----

If we have (2) return:
```
'providers':
- metrics
- health
- oidc
```

Then we would have these URLs available in (3):
`openapi/providers/metrics`
`openapi/providers/health`
`openapi/providers/oidc`

arthurdm:
Fixes #212 

Adding new SPI for platform OpenAPIs.

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
@@ -717,12 +717,13 @@ or any other API the framework may be exposing.
 
 The exposure of these platform OpenAPIs allows for:
 
-*  discovery of available platform endpoints
-*  generation of clients to consume the API (OAS3 -> client_language)
-*  a standard way to document non-application APIs: this is probably the most significant reason 
-for this new SPI, as the number of APIs exposed by frameworks continues to grow and they all need
-to be documented in some way.  We certainly don't want RESTful APIs being documented in HTML tables,
-proprietary / custom formats, etc - if it's a REST API it should be documented using OpenAPI.
+  *  discovery of available platform endpoints
+  *  generation of clients to consume the API (OAS3 -> client_language)
+  *  a standard way to document non-application APIs
+    **  this is probably the most significant reason for this new SPI, as the number of APIs exposed by 
+  frameworks continues to grow and they all need to be documented in some way.  We certainly don't want 
+  RESTful APIs being documented in HTML tables, proprietary / custom formats, etc - if it's a REST API 
+  it should be documented using OpenAPI.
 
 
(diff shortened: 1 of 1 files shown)


jmini:
I am really not in the application server business, but the feature seems reasonable to me...

---

In regard to the other discussion about YAML vs JSON, can the client send an Header `Content-Type: application/json` during the `GET /openapi/platform` call to get the list of component in JSON format?

---

For the embedded platform OpenAPI specs like `GET /openapi/platform/metrics`, what will be their OpenAPI "header", in particular the `servers` bloc? I guess it should be the url of the running server? The `info` block might be dynamic as well...

---

I was also wondering if a merged OpenAPI-spec of all endpoints is desirable or not.

---

While I was reading the document I was telling to myself that the endpoints available under `/openapi/*` could be described as an OpenAPI spec. It would be somehow complementary to the text specification.

arthurdm:
@EricWittmann - any comments on this PR?

EricWittmann:
I'll have a look as soon as I can carve out a bit of time.  Should be today.

Question: this seems like a pretty big addition to the spec for 1.1, especially at this stage.  Perhaps we should defer this to 2.0?

arthurdm:
that's a good point @EricWittmann - I wanted to release `RC1` for `1.1` today and the thought about this SPI being deferred for `2.0` crossed my mind as well.

I guess `1.1` would be then focused on the addition of `PATCH` and all the annotation / model improvements / fixes, which sounds reasonable to me.

arthurdm:
re-targeted to `2.0`

EricWittmann:
+1 - and sorry I didn't get a chance to review this on Friday. :(

PR #308, opened 2018-12-10
Better testing of PATCH in the TCK.  Fixes #291

### LINKED ISSUE #291: Add support for Patch
In EE8 / JAX-RS 2.1 there's now a `Patch` http method: https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/PATCH.html

This issue will integrate that into the MP OpenAPI spec.

EricWittmann:
Apologies for the import re-org noise in this PR.  Also note that the addition of `externalToolBuilders` is for Eclipse - when I re-imported the mp-oai project it installed a Checkstyle plugin.  That worked but it didn't obey the actual ruleset defined in the `pom.xml`.  So I disabled the plugin, resulting in a local `externalToolBuilders` folder being created.  So I'm just excluding it in the .gitignore and in the rat config.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@
 .classpath
 .project
 .factorypath
+.externalToolBuilders/
 bin/
 target/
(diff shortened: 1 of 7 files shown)


jmini:
Thank you, for fixing usage in Eclipse IDE. I have noticed too, that the current setup do not work well.

Issue #209, opened 2018-02-26
The TCK airlines application should produce a valid OpenAPI document.


mrglavas:
Loading the TCK airlines application and running it through OpenLiberty's validator produced the following errors:

- Message: Required "scopes" field is missing or is set to an invalid value, Location: `#/components/securitySchemes/reviewoauth2/flows/authorizationCode`
 - Message: Required "scopes" field is missing or is set to an invalid value, Location: `#/components/securitySchemes/reviewoauth2/flows/password`
 - Message: The encoding property specified cannot be validated because the corresponding schema property is null, Location: `#/paths/~1user~1createWithArray/post/requestBody/content/application~1json`
 - Message: The Path Item Object must contain a valid path. The "GET" operation of the "/user/{username}" path does not define a path parameter that is declared: "username", Location: `#/paths/~1user~1{username}`
 - Message: The Link Object specified an "operationId" field with "getUserByName" value which points to an Operation Object that does not exist, Location: `#/components/links/UserName`
 - Message: The Link Object specified an "operationId" field with "getUserByName" value which points to an Operation Object that does not exist, Location: `#/paths/~1user~1{id}/get/responses/200/links/User name`

mrglavas:
Should verify whether these errors are legitimate by checking the constraints in the OpenAPI specification and then fix them.

arthurdm:
thanks @Rachelwrq

arthurdm:
fixed via PR #315 - thanks @Rachelwrq!

PR #315, opened 2018-12-18
Fixed errors caused by resource files in airline app

Rachelwrq:
After checking the constraints in the OpenAPI specification to verify the errors are indeed legitimate, fixed the errors listed in issue #209
Build the war file for the airline application and tested in Open Liberty server, everything passed.


genie-microprofile:

PR #320, opened 2018-12-20
Automatically hide MP Rest Client interfaces

### LINKED ISSUE #289: Automatically hide MP Rest Client operations
Currently I have the following MP Rest Client interface in my application:

```java
@Dependent
@RegisterRestClient
@Path("/")
public interface PlayerService {

    @GET
    @Path("/player/{playerId}")
    @Produces(MediaType.APPLICATION_JSON)
    @Operation(hidden = true) // hide operation from OpenAPI
    public Player getPlayerById(@PathParam("playerId") String id);

    @POST
    @Path("/rank/{playerId}/recordGame")
    @Operation(hidden = true) // hide operation from OpenAPI
    public void recordGame(@PathParam("playerId") String id, @QueryParam("place") int place, @HeaderParam("Authorization") String token);

}
```

By default, these synthetic MP Rest Client operations show up in the MP OpenAPI web UI, so I have to add `@Operation(hidden = true)` to each operation in order to hide them.  If I do not hide them, they cannot be invoked because they are not actual rest endpoints available on the server, they are client-facing model of what some other server's REST API looks like.

To resolve this issue, MP OpenAPI could check if a class that appears to be a JAX-RS endpoint is an interface or if it is annotated with `@RegisterRestClient` it would be automatically excluded.
Alternatively, we could try to ask MP Rest Client to set the `mp.openapi.scan.exclude.classes` system property for rest interfaces known to the component.

CC @arturdzm @andymc12

arthurdm:
Fixes #289 

- Updates the spec to clarify the behaviour from vendors.  
- Added new TCK test for it.
- Updated list of 1.1 contributors

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,6 @@
         <httpclient.version>4.5.2</httpclient.version>
         <jackson.version>2.9.3</jackson.version>
         <restclient.version>1.1</restclient.version>
-        <cdi.version>2.0</cdi.version>
 
         <!-- Versions of SPEC dependencies  -->
         <asciidoctor-maven.version>1.5.5</asciidoctor-maven.version>
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,8 @@
         <servlet.version>3.1.0</servlet.version>
         <httpclient.version>4.5.2</httpclient.version>
         <jackson.version>2.9.3</jackson.version>
+        <restclient.version>1.1</restclient.version>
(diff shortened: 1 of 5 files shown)


EricWittmann:
Looks like RAT is failing - not sure why.

arthurdm:
Thanks for the review guys.  I removed the `@Dependent` annotation and added the copyright notice to the new class (which made `rat` happy again).

PR #321, opened 2018-12-21
Additional changes to support @Deprecation of ServerVariables

EricWittmann:


arthurdm:
hey @EricWittmann - changes look good, but some of the lines need to be broken up to make the style checker happy:

```
/jobs/genie.microprofile/Open-API-pull-requests/workspace/api/src/main/java/org/eclipse/microprofile/openapi/models/servers/Server.java:113: error: Line is longer than 150 characters (found 167).
/jobs/genie.microprofile/Open-API-pull-requests/workspace/api/src/main/java/org/eclipse/microprofile/openapi/models/servers/ServerVariables.java:29: error: Line is longer than 150 characters (found 165).
```

api/src/main/java/org/eclipse/microprofile/openapi/models/servers/Server.java
```diff
@@ -108,29 +110,70 @@ default Server description(String description) {
...
+     * Variables are represented as a map between variable name and its value. The value is used for substitution in the server's URL template.
+     * </p>
+     * <p>
+     * This method replaces <code>getVariables()</code> (which is deprecated) as part of an effort to eventually remove 
+     * the {@link ServerVariables} class.  A future version of this class will likely transition this method back to 
+     * being named <code>getVariables()</code> once the current method with that name is removed.
+     * </p>
+     * @return ServerVariables variables
+     */
+    Map<String, ServerVariable> getServerVariables();
```
jmini: What is the idea of adding this method?

First I was thinking the other should also be could `setServerVariables(..)` and `serverVariables(..)` do be consistent.

But now I have the feeling that `ServerVariables getVariables()` will changed to `Map<String, ServerVariable> getVariables()` in the future.

Then this new method `Map<String, ServerVariable> getServerVariables();` will no longer be necessary, will be marked as deprecated in `2.0` and will be removed in `3.0`

---
I would let `ServerVariables getVariables();` the way it is in `1.1`, without `@deprecated`. User of the lib can call it and assign the result to `Map<String, ServerVariable>` even if the method is returning a sub-interface: `ServerVariables`.

And remove `Map<String, ServerVariable> getServerVariables()`.

arthurdm: That's a valid point.  Maybe just add in the javadoc for `getVariables` that in future releases this will be changed to `Map<String,ServerVariable>`

EricWittmann: Yeah @jmini you are correct in your analysis.  I went back and forth on this.  But if we do as you suggest I think users will have deprecation warnings when using the method in some IDES.  Even when the method itself isn't deprecated.  That said if you wanted to check that in eclipse and intellij and I'm wrong, then I think we should change it to your suggestion.

EricWittmann: Unfortunately I'm away from a computer for the next week or so...one of you fine gentlemen will need to finish this up.  Sorry about that!

arthurdm:
awesome.  thanks for the great investigation @jmini!

Issue #323, opened 2018-12-29
Behaviour of add methods with null value


jmini:
When a map is used, the model interfaces have `addXxxx(String, T)` methods.

The Javadoc is not clear about `null` beeing accepted or not as value of type T

I think that calling an add method with a `null` value (e.g. `addXxxx(“key”, null)`) is equivalent to calling `removeXxxx(“key”)`. 

---
The test in `ModelConstructionTest` do not enforce this. We could add something like:

```java
Components components = OASFactory.createComponents();

components.addSchema(
        "MyObject", OASFactory.createSchema()
                .type(SchemaType.STRING)
);
// check on components.getSchemas() => it should only contains "MyObject"

components.addSchema("MyObject", null);
// check on components.getSchemas() => it should be empty
```
---

List:
```
Components#addSchema(String, Schema)
Components#addResponse(String, APIResponse)
Components#addParameter(String, Parameter)
Components#addExample(String, Example)
Components#addRequestBody(String, RequestBody)
Components#addHeader(String, Header)
Components#addSecurityScheme(String, SecurityScheme)
Components#addLink(String, Link)
Components#addCallback(String, Callback)

Operation#addCallback(String, Callback)

Paths#addPathItem(String, PathItem)

Callback#addPathItem(String, PathItem)

Header#addExample(String, Example)

Link#addParameter(String, Object)

Content#addMediaType(String, MediaType)

Discriminator#addMapping(String, String)

Encoding#addHeader(String, Header)

MediaType#addExample(String, Example)
MediaType#addEncoding(String, Encoding)

Schema#addProperty(String, Schema)

Parameter#addExample(String, Example)

APIResponse#addHeader(String, Header)
APIResponse#addLink(String, Link)

APIResponses#addAPIResponse(String, APIResponse)

Scopes#addScope(String, String)

SecurityRequirement#addScheme(String, List)

ServerVariables#addServerVariable(String, ServerVariable)
```

---
Example found in a vendor implementation:
```java
    /**
     * @see org.eclipse.microprofile.openapi.models.responses.APIResponses#addAPIResponse(java.lang.String, org.eclipse.microprofile.openapi.models.responses.APIResponse)
     */
    @Override
    public APIResponses addAPIResponse(String name, APIResponse apiResponse) {
        if (apiResponse == null) {
            this.removeAPIResponse(name);
        } else {
            this.put(name, apiResponse);
        }
        return this;
    }
```

jmini:
[...]

PR #324, opened 2019-01-03
Behaviour of add methods with null value

jmini:
See #323

* Improve Javadoc
* Add TCK tests


arthurdm:
thanks @jmini!  great job.

Issue #327, opened 2019-01-04
TCK bugs in Release 1.1


EricWittmann:
I found three problems with the TCK when updating smallrye:

1. The Airlines app test has an assertion using the wrong `operationId` value here:  https://github.com/eclipse/microprofile-open-api/blob/bd7706d3bacfc15e8253978b75348f54df1b232f/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java#L264 The correct value is actually `getUserByName` as found here: https://github.com/eclipse/microprofile-open-api/blob/bd7706d3bacfc15e8253978b75348f54df1b232f/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/UserResource.java#L454

2. The `ModelConstructionTest` checks that null scope descriptions (in the `Scopes` class) are disallowed here:  https://github.com/eclipse/microprofile-open-api/blob/bd7706d3bacfc15e8253978b75348f54df1b232f/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java#L823  However I think null descriptions for scopes are OK and should be allowed.

3. The `ModelConstructionTest` checks that null scopes cannot be added in `SecurityRequirement` in the `addScheme` method here:  https://github.com/eclipse/microprofile-open-api/blob/bd7706d3bacfc15e8253978b75348f54df1b232f/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java#L865  However I think null scopes should be allowed in this case, since the scheme might not support scopes **or** the user might not need any scopes to access the operation(s).

jmini:
As discussed on gitter, I will have a look at 2 and 3

PR #329, opened 2019-01-04
Addressed three problems with the TCK released in v1.1

EricWittmann:
See details in #327 

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
@@ -34,8 +34,8 @@ public interface SecurityRequirement extends Constructible, Map<String, List<Str
      * required scope (optional) provided.
      * 
      * @param securitySchemeName the name of security scheme declared in the Components section of the OpenAPI document
-     * @param scope a required scope - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. null values will be rejected
-     *              (implementation will throw an exception) or ignored.
+     * @param scope a required scope - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. passing null will
+     *              result in an empty list of scopes
      * @return Updated SecurityRequirement instance
      */
     SecurityRequirement addScheme(String securitySchemeName, String scope);
@@ -45,8 +45,8 @@ public interface SecurityRequirement extends Constructible, Map<String, List<Str
      * required scopes (optional) provided.
      * 
      * @param securitySchemeName the name of security scheme declared in the Components section of the OpenAPI document
-     * @param scopes the scopes required - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. null values will be rejected
-     *               (implementation will throw an exception) or ignored.
+     * @param scopes the scopes required - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. passing null
+     *               will result in an empty list of scopes
      * @return Updated SecurityRequirement instance
      */
     SecurityRequirement addScheme(String securitySchemeName, List<String> scopes);
@@ -98,10 +98,10 @@ public interface SecurityRequirement extends Constructible, Map<String, List<Str
     }
 
     /**
-     * Returns a path item for a given name. This is a convenience method for <code>getSchemes().get(name)</code>
+     * Returns a list of scopes for a given scheme name. This is a convenience method for <code>getSchemes().get(name)</code>
      * 
      * @param securitySchemeName the name of security scheme
-     * @return the corresponding path item or null.
+     * @return a list of scopes or null.
      */
     default List<String> getScheme(String securitySchemeName) {
         Map<String, List<String>> map = getSchemes();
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
```diff
@@ -34,8 +34,8 @@
      * required scope (optional) provided.
      * 
      * @param securitySchemeName the name of security scheme declared in the Components section of the OpenAPI document
-     * @param scope a required scope - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. null values will be rejected
-     *              (implementation will throw an exception) or ignored.
+     * @param scope a required scope - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. passing null will
```
arthurdm: since we're updating this line, I would change `when the defined scheme is` -> `when the defined scheme's type is`

api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
```diff
@@ -98,10 +98,10 @@ default boolean hasScheme(String securitySchemeName) {
...
 
     /**
-     * Returns a path item for a given name. This is a convenience method for <code>getSchemes().get(name)</code>
+     * Returns a list of scopes for a given scheme name. This is a convenience method for <code>getSchemes().get(name)</code>
      * 
      * @param securitySchemeName the name of security scheme
-     * @return the corresponding path item or null.
+     * @return a list of scopes or null.
      */
     default List<String> getScheme(String securitySchemeName) {
```
arthurdm: shouldn't this method be called `getScopes`?

jmini: Well `getScheme()` in `SecurityRequirement` is consistent with all the other methods: `addScheme(..)`, `getSchemes()`, ...
I think it is fine the way it is.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
```diff
@@ -862,7 +867,13 @@ public void securityRequirementTest() {
         
         final List<String> otherValue = new ArrayList<String>();
         checkMapImmutable(sr, SecurityRequirement::getSchemes, "otherScheme", otherValue);
-        checkNullValueInAdd(sr::getSchemes, sr::addScheme, "other", otherValue);
+        
+        final String schemeKey3 = "myScheme3";
+        sr.addScheme(schemeKey3, (String) null);
+        assertTrue(sr.hasScheme(schemeKey3), "Expected " + schemeKey3 + " to be present");
+        final List<String> schemeValue3 = Collections.emptyList();
+        assertSame(sr.getScheme(schemeKey3), schemeValue3,
```
jmini: For list comparison, please use `assertEquals(..)`, checking for the same instance is too strong (implementation can use something else than `Collections.emptyList()`)

```suggestion
        assertEquals(sr.getScheme(schemeKey3), schemeValue3, 
```

api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
EricWittmann: I agree with @jmini

EricWittmann:
I've made appropriate updates based on feedback.  Thanks guys.

api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
arthurdm: It's true that it's consistent - and this is 100% fine for the purposes of this PR.  

We may want to take a look for `2.0` on the usage of these methods and if they make sense.  For example, the `schemes` here are simple references to `Scheme` objects defined elsewhere, with the addition of which scopes (if oauth based) apply.  So in `getScheme` we're not getting back the original `Scheme` object, we're just getting the `scopes` (if any) we applied to it.

api/src/main/java/org/eclipse/microprofile/openapi/models/security/SecurityRequirement.java
```diff
@@ -45,8 +45,8 @@
      * required scopes (optional) provided.
      * 
      * @param securitySchemeName the name of security scheme declared in the Components section of the OpenAPI document
-     * @param scopes the scopes required - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. null values will be rejected
-     *               (implementation will throw an exception) or ignored.
+     * @param scopes the scopes required - only valid when the defined scheme is 'oauth2' or 'openIdConnect'. passing null
+     *               will result in an empty list of scopes
```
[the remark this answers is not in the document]
EricWittmann: Done.

EricWittmann:
Thanks for all the feedback - hopefully this PR is now read to merge. :)

EricWittmann:
Note: I have smallrye passing the TCK now, with these changes.

Issue #245, opened 2018-07-24
Models: Are ServerVariables really Extensible?


jmini:
`org.eclipse.microprofile.openapi.models.servers.ServerVariables` represents a map of `org.eclipse.microprofile.openapi.models.servers.ServerVariable` in `org.eclipse.microprofile.openapi.models.servers.Server`

See spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#server-variable-object

I do not think that the map <variable-name, instance> is extensible. I know Swagger-Core has implemented it, but Kaizen Model did not and I think this is correct.

* Swagger-Core: [ServerVariables.java L57-L78](https://github.com/swagger-api/swagger-core/blob/110ae2bae21b7f2e64aff3691953c11a533a14c9/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/servers/ServerVariables.java#L57-L78) 
* KaiZen: [Server.java](https://github.com/RepreZen/KaiZen-OpenApi-Parser/blob/ad8fa48166e0540ba09ed028292a3a06aa512516/kaizen-openapi-parser/src/main/java/com/reprezen/kaizen/oasparser/model3/Server.java#L37) is using `Map<String, ServerVariable>`,  there is no `ServerVariables` interface.

Of course `ServerVariable` is extensible.

jmini:
Thank you for the quick feedback!

Well I guess that removing `extends Extensible` from the `ServerVariables` interface declaration is not a problem:

https://github.com/eclipse/microprofile-open-api/blob/7a28b39dfba5d8546efe5d195f547419f6761ad2/api/src/main/java/org/eclipse/microprofile/openapi/models/servers/ServerVariables.java#L25-L30

I am wondering if this is possible for a `1.1` version (that should stay backward compatible). Maybe I did not get it.

EricWittmann:
I think the `ServerVariables` model should be removed and the `variables` property of `Server` should instead just be a `Map`.  What do you think, @arthurdm ?

jmini:
@EricWittmann : I totally agree, but this is not only a `ServerVariables` question => #248.

arthurdm:
hey guys - I totally agree here.    Probably just legacy stuff we can improve upon.

How about we tackle this in #248, since that will probably eliminate `ServerVariables` along with other similar interfaces?

jmini:
I am not sure anymore about this. Maybe `ServerVariables` is extensible after all.

[...]

PR #330, opened 2019-01-07
Add a test for Server#setVariables(Map<String, ServerVariable>)

jmini:
See #245


EricWittmann:
This looks good to me!  @arthurdm ?

PR #335, opened 2019-02-25
Introduced TCK test base class: OASConfigScanClassBase

EricWittmann:
This base class is for OASConfigScanClassTest and OASConfigWebInfTest to avoid duplicate @Deployment annotations in a test hierarchy.  For why this is a potential problem, see:  https://github.com/thorntail/thorntail/pull/1224


EricWittmann:
I think this approach is better, even if some platforms might pass the TCK as-is.  The reason is that with multiple deployments, the behavior of the system is undefined.  Better to make sure that all TCK tests have a single `@Deployment`.

EricWittmann:
I have confirmed that SmallRye passes the TCK with these changes, so if approved we could do a patch release immediately in my opinion.

PR #336, opened 2019-02-26
Make ModelConstructionTest run within the container

arturdzm:
When running with the container, this test validates that all model classes are accessible from within the application's context (for filter, etc)

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -1,0 +1,35 @@
+//
+// Copyright (c) 2019 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


genie-microprofile:

arthurdm:
sounds good @EricWittmann - please let us know.

jmini:
My use case is to use the interfaces defined in `org.eclipse.microprofile.openapi.models.*` to work with an OpenAPI model, independently from any server or container. 
I have open-sourced a bridge that wraps Swagger-Core into the interfaces and I plan to contribute something similar for KaiZen (see [empoa](https://openapitools.github.io/empoa/))

I know that this is not the primary target of MicroProfile which is a set of spec that comes on top of Ja(va|karta)EE (and on top of JaxRS in the particular case of OpenAPI).

In addition I do not know Arquillian well, but I had the feeling that the ModelConstructionTest do not requires a container.

If you put this back in (reverting https://github.com/eclipse/microprofile-open-api/commit/9e8d57a4b27241cf9b243990fa57be3c8cf306bc), this forces a container to be present. Can I run the tests without any container? If yes how?

See #270 and #274

EricWittmann:
By the way - I think this PR needs to be submitted against the `1.x` branch.  Or when merged it should be cherry-picked.  My last PR was done against `1.x` and I presume our release this week would be done against that branch.  Not sure how you want to handle this, @arthurdm

arthurdm:
good point @EricWittmann.

hey @arturdzm - can we please re-target to the `1.x` branch?  Shouldn't hit any merging errors.

EricWittmann:
@arthurdm @arturdzm - this looks good to me once it's retargetted to the `1.x` branch.  I've tested this with both SmallRye and Thorntail and both now pass the TCK.  So a big +1 from me.

Issue #328, opened 2019-01-04
The Scopes class should be deprecated (similar to ServerVariables)


EricWittmann:
The `Scopes` class seems to be a leftover from OpenAPI 2.0.  The 3.0 specification does not have a `Scopes` object.

OpenAPI 2.0: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#scopesObject
OpenAPI 3.0: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#oauth-flow-object

In 3.0 `scopes` is a simple **Map** property of the `OAuth Flow` object.  As a result it is not extensible and should be modeled as a simple map in MP+OAI.

arthurdm:
I agree - and there's a good example inside `SecurityRequirement` on how to store a scope without needing a new class.

jmini:
This is fixed with PRs:

* #345, targets the 2.0 branch, is allowed to break the API and remove interfaces/methods
* #346, targets the 1.x branch, where you have deprecations in a non-breaking way.

Feel free to reopen if you disagree.

Issue #343, opened 2019-03-25
Create release note on the spec


Emily-Jiang:
The open api spec does not have the release note section. Please add the section and list what changed for 1.1 and 2.0. Please refer to config spec for an example.

arthurdm:
Thanks Emily.  We do have https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/release_notes.asciidoc, but unfortunately it wasn't updated in the 1.1 release.   This workitem should fix that.

jmini:
I think this issue was fixed by #353 and #354. I am closing it now, feel free to reopen it if needed.

Issue #251, opened 2018-08-02
Breaking changes for 2.0


arthurdm:
This issue will encapsulate breaking changes that we're considering for a future `2.0` release.  Things like removing deprecated functionality,  removing incorrect classes / methods, changing parts of the design, etc.   

These items can go into the [2.0 branch](https://github.com/eclipse/microprofile-open-api/tree/2.0).

To add an item to this list please add a comment.  

Current list of `2.0` proposals:

- [ ] update extensible and non-extensible
    Links: [models](https://github.com/eclipse/microprofile-open-api/issues/251#issuecomment-409778777) 
- [ ] Remove methods that were renamed (and deprecated) in 1.1 
    Links: [256](https://github.com/eclipse/microprofile-open-api/pull/256)  [261](https://github.com/eclipse/microprofile-open-api/pull/261) [268](https://github.com/eclipse/microprofile-open-api/pull/268) [280](https://github.com/eclipse/microprofile-open-api/pull/280)
- [ ] Delegate document / property location to MP Config 
    Links: [Arch Board 46](https://github.com/eclipse/microprofile/issues/46)
- [ ] (#240) All getters returning a list or a map MUST return copy of this list/map containing the same items. This list/map CAN be immutable
- [ ] (#284) The setters having a list or a map as parameter MUST not use the list/map instance directly (it can be for example an immutable list/map)
- [ ] (#345) Removal of the `Scopes` mode
- [ ] (#348) Remove ServerVariables
- [ ] (#352) Removed parameter target in Parameters annotation
- [ ] (#384) Add SchemaProperty Annotation

arthurdm:
From #248 (thx to @jmini and @EricWittmann):  
- replacing the non-extensible model interfaces with a map version in the appropriate encapsulating model.
- update extensible model interfaces to not extend from `Map`, and instead just contain a map as an internal field - exposing the non-mutable collection, and appropriate add and remove methods.

jmini:
Will you create a new branch to start to work on this version? This way PRs can be opened to start to work on this.

arthurdm:
Yup, good point.  Created: https://github.com/eclipse/microprofile-open-api/tree/2.0

jmini:
Remove methods that were renamed (and deprecated) in 1.1: #256

arthurdm:
thx - added to list

jmini:
For this point:
> Remove methods that were renamed (and deprecated) in 1.1

Add: #261

[...]

PR #345, opened 2019-04-12
Remove Scope

Rachelwrq:
For issue #328 deprecated `Scope` class to fit OpenAPI 3.0, now `scope` is just a simple **Map** of strings.


genie-microprofile:

jmini:
@Rachelwrq thank you for this patch.

I will review more in depth, but I have the feeling that with the way this PR is written, `Scopes` can't be used anymore. Existing code will break (`Scopes` could be completely removed, it is no longer used in the model at all) and there was no previous released version indicating that this will be the case.

---

For the moment all the breaking changes introduced with `2.0` are marked as deprecated in `1.1`. If you remove all those warnings with `1.1` (and switch to the new way of expressing stuff) and then switch to `2.0` it will work.

I was thinking that this is the rule in Microprofile projects.

--- 
As a reference have a look on how @EricWittmann prepared the removal of `ServerVariables` in `1.1`:
* b459d3d20ddb60d2c009727769568831ff1e6f18
* f7970b9d0787bbd7f70f06fa7a52973b16ca1f33
* 2dbe14058058bbd831dba8a24dd48d0d8e07c727

arthurdm:
hi @jmini - I am wondering if since now master == `2.0`,  would we continue to update the `1.x` branch?

Rachelwrq:
@jmini Thank you for looking into my PR.

I made some modifications in reference to Eric's commits as well as the most recent files relating to `ServerVariables`; so that the `Scopes` is properly deprecated by now. 

Please let me know if there is anything else I might have missed.

jmini:
> I am wondering if since now master == 2.0, would we continue to update the 1.x branch?

@arthurdm I do not think so.

In my opinion, `1.x` is the maintenance branch for the `1.1.x` versions (if something is found that requires to build again a `1.1.x` release)

jmini:
@Rachelwrq I will. 

I would like to test it with my implementation, but I am a little behind with catching up with the changes on master. I hope you do not mind the delay...

arthurdm:
> In my opinion, 1.x is the maintenance branch for the 1.1.x versions (if something is found that requires to build again a 1.1.x release)

So with that thought, do we really need to deprecate things?  I was thinking, until `2.0` is not released, we can do breaking changes in master?

jmini:
> we can do breaking changes in master?

You are the MicroProfile expert, I have no idea of the rules... Should I ask the root project for opinion?

In my eyes there are 2 types of breaking changes (I just made up the names, maybe there is an official terminology). Applied to removal:

* A) Hard breaking change: you just do not care about backward compatibility. You are allowed to remove what you want when you introduce a new major release.
* B) Soft breaking change: you are only allowed to remove something in a major version that was marked as deprecated in a previous release (and some projects add a notion of time: the item being removed must be marked as deprecated during X releases or during X months).

Java SE is clearly type B. The eclipse platform (Eclipse RPC, IDE - which has nothing to do with MicroProfile) is also type B.

If you tell me than type A is allowed here, it makes our life easier!

pilhuhn:
@arthurdm In Metrics we have iirc deprecated one thing in 1.x (if at all) and are making a lot of breaking changed in MP-Metrics 2.0. Those breaking changes are the reason we did not hurry at the end to get into MicroProfile 2.2, as those minor releases of the umbrella are not meant to have breaking changes. 
tl;dr: I'd say as long as you release a 2.0 of OpenAPI and 2.0 goes into MP-3.0, you should be fine.

arthurdm:
thanks @pilhuhn!

arthurdm:
hey @Rachelwrq - sorry for taking long to circle around this issue.  We had a chat at the last hangout and decided that:

- this PR, as it targets the `2.0` branch, is allowed to break the API and remove interfaces/methods
- a new PR should be created, to target the `1.x` branch, where you have deprecations in a non-breaking way.

Rachelwrq:
Hi @arthurdm 
I’m just wondering if it is okay for me to just edit this PR so it is pushed to the `1.x` branch since I made all my changes to properly deprecate `Scope` object. I can then introduce breaking changes and delete interface and method in another PR.

jmini:
@arthurdm: Thank you a lot for summarizing the hangout discussion. I like the outcome.

jmini:
I do not remember where (Gitter?), but we were told to have an document or a section listing all the changes with the next version. This comment https://github.com/eclipse/microprofile-open-api/issues/251#issuecomment-481403906 is also mentioning it, but it does not tell where to put this.

arthurdm:
@jmini - I believe you're looking for something [like this](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/release_notes.asciidoc)?    definitely a good point.

jmini:
@arthurdm this means we do not have something like that in Microprofile-OpenAPI ?

arthurdm:
My mistake, I used the wrong link.  We do have it [here](https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/release_notes.asciidoc), but as per https://github.com/eclipse/microprofile-open-api/issues/343#issuecomment-476318978 we missed the update in `1.1`

arthurdm:
added to the list.  thanks!

PR #346, opened 2019-04-30
Deprecate Scope Object in 1.x

Rachelwrq:
This PR is for issue #345 
Deprecated the `Scope` object and use simple maps to represent scopes in `OAuthFlow`


genie-microprofile:

arthurdm:
should the deprecating notice say `since 1.2`? (instead of `1.1`)

Rachelwrq:
Hi @arthurdm 
My bad. I changed all of the `Scopes` related `1.1` to `1.2`.

arthurdm:
ya, probably just `2.0`.  But we'll probably have to wait until the next `breaking change` umbrella release, which will take sometime.    So if there's a need we could do `1.2`.

PR #348, opened 2019-05-17
Remove ServerVariables

jmini:
Issue: #245
This is a follow-up of PRs: #317, #321, #330

---

With `1.1` we have indicated that `ServerVariables` will be removed with `2.0` and we have told our user they should use `Map<String, ServerVariable>` instead of `ServerVariables`. Back then `ServerVariables` was a subtype of `Map<String, ServerVariable>`

Currently with `2.0-SNAPSHOT`, `ServerVariables` is no longer a subtype of Map<String, ServerVariable>`, this will break user code that have followed our advice and that have started to use `Map<String, ServerVariable>`.

This PR is doing the rest of the work as planned in #245, removing `ServerVariables` completely, (this is similar to what we just did in #345 for `Scope`).

---
cc @EricWittmann, @arthurdm


api/src/main/java/org/eclipse/microprofile/openapi/models/servers/Server.java
```diff
@@ -111,19 +111,24 @@ default Server description(String description) {
      * </p> 
      * @return ServerVariables variables (in the future this method will instead return a <code>Map&lt;String, ServerVariable&gt;</code>).
```
arthurdm: can we update this line?  Perhaps also mention that changes to the returned Map aren't guaranteed to be reflected in the model?

jmini: Thank you for catching this, I have updated my commit.

Issue #340, opened 2019-03-18
Parameters annotation should only target the method


arthurdm:
Current our [Parameters annotation](https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/parameters/Parameters.java#L33) has two targets: method and parameter.

@arturdzm brought up a good point that it should only have the method target.  Otherwise, if we associate `@Parameters` with a single parameter, we're associating `multiple` parameters with a single one..which doesn't make much sense.

arthurdm:
PR https://github.com/eclipse/microprofile-open-api/pull/352 has completed the development work piece of this issue.  However, we're still missing the corresponding TCK update.  At the moment the TCK should not complile.

EricWittmann:
@arthurdm This seems complete now with https://github.com/eclipse/microprofile-open-api/pull/358 ?

arthurdm:
agreed.  closing.

PR #352, opened 2019-05-23
Removed parameter target in Parameters annotation

Rachelwrq:
This PR fixes issue #340 
With the parameter target removed, `@Parameters` now only have the method target.


genie-microprofile:

PR #358, opened 2019-06-17
adjusted tck parameters notation

Rachelwrq:
This PR addresses the comment in issue #340 
Updated the Parameters notation inside of airlines resources so that TCK can compile


genie-microprofile:

PR #379, opened 2019-10-22
Code improvements about semantics and best practices

rhuan080:
Signed-off-by: Rhuan Rocha <rhuan080@gmail.com>

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/data/PetData.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/data/PetData.java
@@ -13,17 +13,20 @@
 
 package org.eclipse.microprofile.openapi.apps.petstore.data;
 
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.openapi.apps.petstore.model.Category;
 import org.eclipse.microprofile.openapi.apps.petstore.model.Pet;
 import org.eclipse.microprofile.openapi.apps.petstore.model.Tag;
 
 import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Objects;
+import java.util.Map;
+
 
 public class PetData {
-  static List<Pet> pets = new ArrayList<Pet>();
+  static List<Pet> pets = new ArrayList<>();
   static List<Category> categories = new ArrayList<Category>();
 
   static {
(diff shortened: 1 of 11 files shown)


genie-microprofile:

genie-microprofile:

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/data/PetData.java
```diff
@@ -13,14 +13,12 @@
...
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.microprofile.openapi.apps.petstore.model.Category;
 import org.eclipse.microprofile.openapi.apps.petstore.model.Pet;
 import org.eclipse.microprofile.openapi.apps.petstore.model.Tag;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.HashMap;
+import java.util.*;
```
arthurdm: do we really need to import everything from that package?

EricWittmann: Definitely not. :)

rhuan080: Hi, I'm so sorry. It is not needed.

rhuan080: I'll fix it.

rhuan080: Updated.

MikeEdgar:
I would personally avoid using the `StringUtils` `isBlank` and `isNotBlank` methods for a couple of reasons. First, because it increases the direct dependency on commons-lang3 for a trivial piece of functionality. Second, because it is logically different than the code it's replacing in this case.

* `status != null && !"".equals(status)` is only true if the string is non-zero length (i.e. whitespace would be allowed)
* `StringUtils.isNotBlank(status)` is true only for non-whitespace characters.

I would suggest using `isEmpty` on the `String` instance after the `null` check instead of the `StringUtils` method calls.

rhuan080:
--> Second, because it is logically different than the code it's replacing in this case.
In this code snippet the status should not be a whitespace. 

--> First, because it increases the direct dependency on commons-lang3 for a trivial piece of functionality.
I think it is not a problem because commons-lang3 is a well known lib that is used a lot.

Issue #360, opened 2019-07-08
Properties can not be defined in @Schema


jmini:
@agiertli asked on Gitter (Jun 24 15:28) following question:

----

Hi folks,

I am trying to annotate a parameter which holds an uploaded file in my POST request:

The final OpenAPI spec should look like this:
https://swagger.io/docs/specification/describing-request-body/file-upload/

```yaml
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                fileName:
                  type: string
                  format: binary
```

I can't seem to find a suitable way how to define 'properties'., when using MicroProfile annotations for OpenAPI. So far, I have this:

```java
public Response processInvoice(
        @RequestBody(content = @Content(mediaType = "multipart/form-data", schema = @Schema(type = SchemaType.OBJECT))) MultipartFormDataInput input)
```

Which generates:

```yaml
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
```

So it seems I am on the right track, but not sure how to proceed further..any ideas?

jmini:
My idea back then was:

Yes it seems to me that you are right the [`@Schema`](https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java) annotation do not let you define any properties as member.

A way to achive this, could be that you create a reference to an other object.

```yaml
      requestBody:
        content:
          multipart/form-data:
            schema:
                $ref: '#/components/schemas/MyRequestBody'
```

With a component section being like this:

```yaml
components:
  schemas:
    MyRequestBody:
      description: TODO
      type: object
      properties:
        fileName:
          type: string
          format: binary
```

From an OpenAPI point of view the indirection over the components section is semantically the same.

And then to create the `MyRequestBody` object in the schemas section of the components section:

* either you have it defined in a static file (without the paths that will be populated by annotation scanning)
* or you implement a `OASModelReader` that create your additional schemas in the `buildModel() method

Instead of sharing the schema in the component section you can even share the complete "request body" section

---

I think `@Schema` could provide a way to set properties directly.

EricWittmann:
I think that a fairly simple solution to this (and perhaps one that is not controversial) would be to introduce `properties` to the `@Schema` annotation.  So for example, an application of the `@Schema` annotation might look like this:

```java
@Schema(properties={
    @Schema(name="creditCard", required=true),
[...]

PR #384, opened 2019-11-06
Add SchemaProperty Annotation (#360)

MikeEdgar:
This is an initial attempt at implementing #360. The new `@SchemaProperty` annotation added is different from `@Schema` by not having a `properties` attribute and also by not specifying a `Target` (i.e. it can only be used from `@Schema`).

Please review the JavaDoc on `@Schema#properties()`. It was worded that way for "ease of use", but it may also be useful to specify that it behaves some other way. 

Based on the outcome of the review, TCK scenarios can be coded.


api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/SchemaProperty.java
```diff
@@ -0,0 +1,371 @@
+/**
+ * Copyright (c) 2017 Contributors to the Eclipse Foundation
+ * Copyright 2017 SmartBear Software
```
arthurdm: we probably don't need to include the SmartBear copyright for new file?

arthurdm:
Thanks @MikeEdgar!    Would it helpful to include a sample in the javadoc? 

Something like this, from #360
```java
@Schema(properties={
    @SchemaProperty(name="creditCard", required=true),
    @SchemaProperty(name="departureFlight", description="The departure flight information."),
    @SchemaProperty(name="returningFlight")
})
```

api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/SchemaProperty.java
MikeEdgar: Removed - I wasn't sure about this one since the file was largely a clone of Schema.java.

arthurdm:
thanks @MikeEdgar.  Should we also get rid of minProperties, maxProperties and requiredProperties  from `@SchemaProperty`- or could those be valid if the type of the sub-property points to a POJO implementation?

MikeEdgar:
@arthurdm - yes I think those could be used in `@SchemaProperty`, even in the absence of a POJO, such as when a generic object type is modeled. 

One option that is questionable is `required` since the parent `@Schema` also has `requiredProperties`. However, I hesitate to remove that in case there are scenarios where merging POJO annotations with `@SchemaProperty` is more flexible with both options.

MikeEdgar:
@arthurdm - what are your thoughts on keeping/removing `required` in `@SchemaProperty`? If the current state of the annotation is good to go, I'll update this PR with TCK tests.

arthurdm:
hey @MikeEdgar - ya, perhaps we should remove the `required` field from `@SchemaProperty`, otherwise we'll need to document (and implementations handle) the case where this field disagrees with the parent's list of `requiredProperties`.

MikeEdgar:
Updated pushed, along with a simple TCK test. Note that the `User` model had both a `userName` field and a `getUsername` method, which slightly disagree in capitalization. This PR includes renaming the getter/setter to align with the field.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
```diff
@@ -287,7 +287,7 @@
      * 
      * @return the type of this schema
      **/
-    SchemaType type() default SchemaType.DEFAULT;;
+    SchemaType type() default SchemaType.DEFAULT;
```
EricWittmann: Nice. :)

arthurdm:
merged.  thanks @MikeEdgar!

PR #386, opened 2019-11-19
Add TCK test for checking enum schema

Cousjava:


MikeEdgar:
Would it make sense to have the enumerated string values be slightly different than the Java enum constants to make sure the implementation uses the annotation? E.g. lower case of the same names.

EricWittmann:
While we're making changes here - would anyone else feel slightly more comfortable if we changed it from `Sex` to `Gender`?  :)

Cousjava:
@MikeEdgar @EricWittmann  done

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -650,6 +650,7 @@ public void testSchema(String type) {
         vr.body("paths.'/user'.post.requestBody.content.'application/json'.schema.maxProperties", equalTo(1024));
         vr.body("paths.'/user'.post.requestBody.content.'application/json'.schema.minProperties", equalTo(1));
         vr.body("components.schemas.User.required", hasItems("id", "username", "password")); // requiredProperties
+        vr.body("components.schemas.Sex.enum", hasItems("Male", "Female", "Other"));
```
EricWittmann: Does this still work after the Sex->Gender rename?  Otherwise looks great to me. :)

Cousjava: Done

PR #403, opened 2020-01-11
Schema Registration Using Config Property

### LINKED ISSUE #364: Provide mechanisms for users to register own types and format
As discussed in the latest hangout (and related to https://github.com/smallrye/smallrye-open-api/issues/147)

It would be useful to have a mechanism that allows users to:

- Provide their own mappings for types (whether their own, or overriding defaults) to particular primitive type and formats -- see Mike Edgar's comments here https://github.com/smallrye/smallrye-open-api/issues/147#issuecomment-507232015

For example, this could be as simple as:
 `mp.openapi.extensions.type-format.FQCN=SchemaType,DataFormat`

So you could map your custom date type to the correct SchemaType and DataFormat.

- As suggested by @EricWittmann in the the meeting was that we may also want to consider a more advanced mechanism that allows a schema to be mapped to a particular object (i.e. for cases where it's not simply about primitives). Eric mentioned the recent SmallRye Custom Schema Registry (https://github.com/smallrye/smallrye-open-api/pull/126/files) as a possible model.

- I suggested that (in addition to config-driven), a programmatic interface might be beneficial, in allowing runtimes to inject certain mapping information they already know about (e.g. for some type that a framework commonly provides).

There are some parallels here with the way that Jackson works -- they have a large number of converters OOTB that take (e.g. Map, Simple Java Bean, List, Date) and convert it into something sensible in JSON. However, if you provide your own complex data structure you need to provide your own converter and/or mapping.

MikeEdgar:
Fixes #364 

Add documentation and TCK test for `mp.openapi.schema.` configuration. Note that the list of property keys has changed from a table to a list to help readability.

Signed-off-by: Michael Edgar <michael@xlate.io>

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.adoc
@@ -93,29 +93,64 @@ https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/o
 
 The following is a list of configuration values that every vendor must support.
 
-[cols="1,4"]
-|===
-| Config key | Value description
+`mp.openapi.model.reader`::
+Configuration property to specify the fully qualified name of the <<OASModelReader>> implementation.
 
-| `mp.openapi.model.reader` | Configuration property to specify the fully qualified name of the <<OASModelReader>> implementation.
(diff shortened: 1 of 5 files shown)


EricWittmann:
I think this is ready to go, @arthurdm

arthurdm:
looks great!  thank you @MikeEdgar

PR #410, opened 2020-02-19
Add RequestBodySchema and APIResponseSchema annotations

### LINKED ISSUE #363: Streamline annotating a method's @Response when using JAX-RS Response object
Originally posted in SmallRye here:  https://github.com/smallrye/smallrye-open-api/issues/76

Content copy/pasted from source:

---

OpenApi has an [ApiResponse annotation](https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponse.java) which, via its nested Content, can provide the information about the JAX-RS Response entity class.

While this annotation covers well for a case where Response dynamically sets different entity types per the current response media type, it is too complex for a typical case of Response returning a known single entity type only, with @Produces already typing a response media type.

One proposal is to have something like this:
```java
@Produces("application/json")
@ApiResponseType(MyType.class)
@GET
public Response get() { return Response.ok(new MyType()).build(); }   
```
but other options should also be considered

MikeEdgar:
Fixes #363 

The changes here are the latest mentioned in #363, with one difference being that the `responseDescription` property of `@APIResponseSchema` is not defaulted to an empty string. This seems appropriate given the JavaDoc of `@APIResponse#description()`:

> A short description of the response. It is a REQUIRED property unless this is only a reference to a response instance.

@EricWittmann @arthurdm - this can be discussed in the next hangout or here.

Signed-off-by: Michael Edgar <michael@xlate.io>


MikeEdgar:
We discussed in the hangout that the documentation for the response annotation will be updated to specify how implementations should derive the `responseCode` and `responseDescription` when they are omitted by an application. Both properties will be updated to be optional (i.e. defaulted to `""`).

MikeEdgar:
Update - I'm working on TCK tests for these annotations as well. Those can be part of this PR or a separate issue - either way is fine with me.

EricWittmann:
I have no preference either - whatever is easier for you I would say.

MikeEdgar:
This branch was re-based and the last two commits with the TCK tests were added.

@EricWittmann , @arthurdm - I think this is ready to go now. Please let me know if you have any feedback.

MikeEdgar:
@arthurdm - have you had a chance to review? Please let me know if you notice any issues.

Issue #404, opened 2020-01-12
Add response code to all endpoint when there are ExceptionMapper(s) defined


FreifeldRoyi:
Hi
When adding annotations to all the endpoints I expose, it gets really cumbersome to copy response annotations that are defined in ExceptionMapper classes

I thought it would be better to annotate the ExceptionMapper definition and propagate it to all endpoints automatically

EricWittmann:
This is actually being worked on right now in SmallRye OpenAPI.  Do you think a change to the spec is needed for this to be supported?  Right now it's not something specifically called out by the spec, which leaves it up to each implementation.  Or we could update the SDK to cover this.

Which implementation are you currently using?

FreifeldRoyi:
I must say that I'm not too familiar with the spec so I'll just shoot whatever comes to mind. I guess that `@Schema` (or something equivalent Say... `@ErrorCode`) needs to be added as a tag. You'll probably want to give it a name in order to add something like `ignore` param, maybe to `@ApiResponse`, (with a list of names) since not all REST calls will need the entire mapping sheet.
Or you can go with the less pessimistic flow and add `errorCodeNames` (instead of `ignore`) to incorporate the relevant mappings. The latter actually has a nice benefit since it's very clear when you read the code, instead of searching for ExceptionMapper(s) all around

SmallEye (Quarkus) or Quarkus(SmallEye) =)

EricWittmann:
Thoughts @MikeEdgar and @lamtrhieu - how does this compare to what's being worked on in SmallRye already?

MikeEdgar:
The main difference here is the linkage between the resource method and the `ExceptionMapper`. The changes in SmallRye use the method's `throws` to link the two rather than annotations. If the resource method declares the same `responseCode` that is present in the mapper, the resource response is the one that is used. We decided against doing anything with undeclared + unchecked exceptions for now.

EricWittmann:
Do you think there are any existing MP annotations that could be supported on Exception Mappers to improve support?

Note: I'm not against inventing new annotations in the spec to cover this use case if it's useful to do so. Which is why I marked this as a discussion topic for the next hangout.

MikeEdgar:
Nothing jumps out at me beyond the `@APIResponse`. There may be room to modify/add an annotation to explicitly provide for the linkage between resource methods and exception mappers. 

```java
@APIResponses(
    exceptionMappers = { My404Mapper.class, My500Mapper.class },
[...]

PR #414, opened 2020-03-03
Updated javadoc and TCK to support @APIResponse on ExceptionMappers

EricWittmann:
This is a fix for #404


EricWittmann:
Note to @MikeEdgar - I already tested this with SmallRye `master` and it works without needing any change to SR.

FreifeldRoyi:
WOOHOO!

EricWittmann:
> btw - coincidence that this fixes issue #404 and you have chosen an example that returns code 404? 😄

Coincidence, yes - but a happy one!

PR #418, opened 2020-04-14
Add general Number comparison Matcher, update pet store assertions

### LINKED ISSUE #412: TCK bug? Use of equalTo PetStoreAppTest#testSchema rejects 10.0 for schema.multipleOf
In `PetStoreAppTest#testSchema`, line 58 retrieves `schema.multipleOf` and uses Hamcrest's `equalTo(10)` to check the value returned from the server under test.

The `Schema` annotation declares `multipleOf` as a `double`, and the `petId` parameter on  `PetResource#getPetById` is 10. 

The `Schema` model interface defines 'multipleOf' as a `BigDecimal`.  As it turns out the SnakeYAML serializer emits `10.0` in this case, using the `BigDecimal` type on the model interface to drive the output format. 

This causes the TCK test to fail -- incorrectly I think -- because the test uses Hamcrest's `equalTo(10)` which does not match `10.0`.

I have worked around this in the extensions we use with SnakeYAML so we can pass that TCK test, but it's a bit of a pain.

(There is a mention of the `double`/`BigDecimal` mismatch regarding `multipleOf` in #178 (closed) but that issue is really not directly related to this one.) 

Using decimal notation to format a `BigDecimal` seems perfectly reasonable, so can the test be changed to accept `10` and `10.0`? 

I have not looked for similar constructs elsewhere in the tests that might also need attention. This was the only test failure I saw due to this problem.

MikeEdgar:
Fixes #412

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
@@ -17,12 +17,13 @@
 package org.eclipse.microprofile.openapi.tck;
 
 import static io.restassured.RestAssured.given;
+import static org.eclipse.microprofile.openapi.tck.utils.TCKMatchers.comparesEqualToNumber;
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.equalTo;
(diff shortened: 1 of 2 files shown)


MikeEdgar:
@tjquinno, FYI

PR #457, opened 2020-12-07
fixing build issue due to wrong TCK dependency version

brunocrt:
commit 0352d7a1f4da489fbbe1aece72305d3a102ba174
Author: Bruno Tinoco <brunocrt@gmail.com>
Date:   Mon Dec 07 12:47:00 2020 GMT-3

    [458] Build Failing on TCK due to missing Rest Client version (2.0) on Maven Central
    
    When multiple projects are imported together, perform all the necessary
    auto shares in a single job rather than spawning a separate job for each
    project.
    
    Bug: https://github.com/eclipse/microprofile-open-api/issues/458
    Signed-off-by: Bruno Tinoco <brunocrt@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <java-hamcrest.version>2.0.0.0</java-hamcrest.version>
         <httpclient.version>4.5.2</httpclient.version>
         <jackson.version>2.10.1</jackson.version>
-        <restclient.version>2.0</restclient.version>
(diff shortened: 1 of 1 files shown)


phillip-kruger:
Hi @brunocrt - thanks for this. We know about this. We are working on a solution to be able to build from the staging repo in pas nexus. But until then, the build will fail for now. We need to stage a version 2.0 of openapi that use config 2.0, both will be in pas nexus staging

PR #470, opened 2021-02-11
Remove broken ClassLoader delegation

stuartwdouglas:
It is up to the ClassLoader to do parent delegation,
we should not be second guessing it by attempting to
manually delegate to the parent. This manual
delegation breaks isolated class loading schemes.

Signed-off-by: Stuart Douglas <stuart.w.douglas@gmail.com>


stuartwdouglas:
This has been an issue with a few other specs as well, e.g. https://github.com/eclipse/microprofile-reactive-streams-operators/pull/130/commits/4e78d38cb26c1a5617e3107916642d5744523b2d

phillip-kruger (approved):
Thanks @stuartwdouglas !

PR #471, opened 2021-02-16
use MediaType.APPLICATION_JSON.

lucguinchard:
replace `"application/json"` to `MediaType.APPLICATION_JSON`


phillip-kruger (approved):
Thanks :)

PR #478, opened 2021-04-15
<fix>(417): Changed applicable context for ApiResponse and ApiResponses annotations

### LINKED ISSUE #417: APIResponse and APIResponses should target TYPE as well
Currently APIResponse and APIResponses use:
`@Target({ ElementType.METHOD })`

The Swagger equivalent handles this differently (see
https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-annotations/src/main/java/io/swagger/v3/oas/annotations/responses/ApiResponses.java):
`@Target({METHOD, TYPE, ANNOTATION_TYPE})`

This makes it possible to add a response to **all** `@Operation`s in a class. It would be great if this was possible with MP OpenAPI as well!

Name-not-available:
I really want this to be done in order to highly decrease amount of boilerplate in my project.

Fixes #417.

Thank you.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponse.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponse.java
@@ -68,7 +68,7 @@ import org.eclipse.microprofile.openapi.annotations.media.Content;
  * @see "https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#responseObject"
  * 
  **/
-@Target({ ElementType.METHOD })
+@Target({ ElementType.METHOD, ElementType.TYPE, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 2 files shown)


nikita-kuklin:
Will this be merged after legal agreement signment?

Emily-Jiang:
> Will this be merged after legal agreement signment?

TCKs for this need to be added as well.

MikeEdgar:
@Name-not-available are you interested in picking this back up? I think it might be good to limit this change to just adding `ElementType.TYPE`. We could revisit adding `ElementType.ANNOTATION_TYPE` in the future when that can be added consistently to any other annotations necessary in this project. 

If you'd like to work this one, please take a look at updating the TCK so that (1) at least one of the test resources has an `@APIResponse` at the class level and (2) the tests are updated to check that the response is applied to the operations contained in the resource class. Of course you'll need to complete the necessary legal agreement as well.

MikeEdgar:
Replaced by #524

PR #488, opened 2021-08-13
Use entry identity in list immutability check

### LINKED ISSUE #453: TCK Tag Collection Test contains() side effect
This line is supposed to check if a list is immutable:

https://github.com/eclipse/microprofile-open-api/blob/b7a437d9efd7ad113895ceb4ba8a1f10614fd779/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java#L421-L422

The test method adds a few default `Tag` objects beforehand, and this one has no unique data. This means that if the OpenAPI implementation properly implements `equals()` and`hashCode()` methods it will actually fail this test, as even if the list is immutable it will contain the previously added Tag (with identical fields).

Is this intended behaviour?

MikeEdgar:
Fixes #453 

The intent of the TCK helper method `checkListImmutable` is to assert that the third argument `otherValue` fails to be inserted to the list returned by the second argument `listGetter`. The current implementation utilizes `List#contains` for this assertion, but `contains` relies on `equals` whereas the test only needs to check object identity. Tests utilizing this method make no assertion regarding a vendor's implementation of `equals` on the objects in the list.

### DIFF
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -33,6 +33,7 @@ There are no functional changes introduced in this release, except the dependenc
 
 * Negative Test Scenario - @SchemaProperty Precedence Behaviour (link:https://github.com/eclipse/microprofile-open-api/issues/466[466])
 * Use MediaType.APPLICATION_JSON instead of application/json in some TCKs (link:https://github.com/eclipse/microprofile-open-api/pull/471[471])
+* TCK Tag Collection Test contains() side effect (link:https://github.com/eclipse/microprofile-open-api/issues/453[453])
 
 [[release_notes_20]]
 == Release Notes for Microprofile OpenAPI 2.0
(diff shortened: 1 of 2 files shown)


Emily-Jiang:
@MikeEdgar can you resolve the conflict and then add this issue to the release note so that we can merge into master for this upcoming release?

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -33,6 +33,7 @@ There are no functional changes introduced in this release, except the dependenc
 
 * Negative Test Scenario - @SchemaProperty Precedence Behaviour (link:https://github.com/eclipse/microprofile-open-api/issues/466[466])
 * Use MediaType.APPLICATION_JSON instead of application/json in some TCKs (link:https://github.com/eclipse/microprofile-open-api/pull/471[471])
+* Use entry identity in list immutability check (TCK) (link:https://github.com/eclipse/microprofile-open-api/pull/488[488])
```
Emily-Jiang, line 36: @MikeEdgar Can you use issue number #453 instead? We try to stick with issue if possible.

MikeEdgar: @Emily-Jiang , updated the link. Thanks

Emily-Jiang (approved):
LGTM! Thank you @MikeEdgar !

PR #495, opened 2021-10-05
Use fixed assertions

### LINKED ISSUE #494: TestNG 7.4.0 Assert.assertNotSame has a bug which causes ModelConstructionTest TCK to fail
assertNotSame ought to assert that two objects are not the same (==) but in version 7.4.0 of TestNG it tests if they are equal. This can cause a TCK failure where two object are expected to be different instances even though they are equal. The tests that fail for me are
- ModelConstructionTest.openAPITest
- ModelConstructionTest.operationTest
- ModelConstructionTest.securityRequirementTest
These all use `ModelConstructionTest.createConstructibleInstance` which uses `Assert.assertNotSame`
https://github.com/eclipse/microprofile-open-api/blob/d768dbe6e267d4b37e09c6c676474cd9d7786208/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java#L1379

The bug was reported multiple times in TestNG and has been fixed but is not yet in a released version.
https://github.com/cbeust/testng/issues/2486
https://github.com/cbeust/testng/issues/2616

Although this issue has been found in the MP OpenAPI TCK, it may be a problem across the whole of MicroProfile 5.0, where the TestNG version is 7.4.0. I suggest that for now the MicroProfile TCKs should avoid use of Assert.assertNotSame.

tevans78:
Signed-off-by: tevans <tevans@uk.ibm.com>

assertSame and assertNotSame are broken in TestNG 7.4.0 - https://github.com/cbeust/testng/issues/2486
This PR adds the same methods, but instead re-implemented using Hamcrest Matchers
Should be removed when TestNG 7.5.0 is released

fixes #494

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
@@ -16,12 +16,16 @@
 
 package org.eclipse.microprofile.openapi.tck;
 
+// assertSame and assertNotSame are broken in TestNG 7.4.0 - https://github.com/cbeust/testng/issues/2486
+// import static org.testng.Assert.assertNotSame;
+// import static org.testng.Assert.assertSame;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.sameInstance;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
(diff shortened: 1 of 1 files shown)


MikeEdgarCHAMP (approved):
LGTM, thank you.

MikeEdgar (approved):
LGTM, thank you.

MikeEdgar:
@Emily-Jiang - is this needed for the jakarta branch?

Emily-Jiang:
Thank you @MikeEdgar for your timely review!

PR #512, opened 2022-02-26
Support Schema `additionalProperties` via annotation

### LINKED ISSUE #423: Support defining "additionalProperties" in @Schema annotation
With #360 support was added for `properties`, but I can't find any support for `additonalProperties` (see https://swagger.io/docs/specification/data-models/dictionaries/). AFAICT it looks like the model does support additional properties (see also #100), but I cannot find anything in the `@Schema` annotation.

MikeEdgar:
Fixes #423


MikeEdgar:
Let's discuss this on the next call. I took a slightly different approach here than last described in #423. By adding support for two special cases of a `Schema` (rendered as `true` or `false`) this simplifies the specification of `additionalProperties` and also can be leveraged in the future with OpenAPI version 3.1.0+ which use JSON Schema directly. In JSON schema, any schema can be specified as a boolean to match everything (`true`) or nothing (`false`).

api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
```diff
@@ -399,4 +399,31 @@
...
+     * {@linkplain SchemaType#OBJECT OBJECT}, either explicitly or as derived by the placement of the annotation.
+     * 
+     * @return a class that describes the allowable schema for additional properties not explicitly defined
+     * 
+     * @since 3.1
+     * 
+     * @see org.eclipse.microprofile.openapi.models.media.Schema.TrueSchema
+     * @see org.eclipse.microprofile.openapi.models.media.Schema.FalseSchema
+     */
+    Class<?> additionalProperties() default org.eclipse.microprofile.openapi.models.media.Schema.TrueSchema.class;
```
Azquelt, line 427: It seems odd that `properties` is a `SchemaProperty` while `additionalProperties` is a `Class<?>` when they seem to document very similar things.

MikeEdgar: Just to add, most of the other attribute that specify a "remote" schema utilize the `Class<?>` approach. `properties` is a special case to allow the fields of an object-type to be explicitly listed out in-line, allowing overrides.

https://github.com/eclipse/microprofile-open-api/blob/2f6592e609489c7dc47eb8698f63d86d7e8ef27c/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/JAXRSApp.java#L78-L79

Azquelt:
Hi, sorry it's taken me a while to get back to this. I've had another think about it.

1. I'd forgotten how severe the restrictions on annotations are - particularly that you can't have an annotation refer to itself. With that in mind, it makes much more sense that we use the `Class<?>` approach to reference nested schemas. It's not ideal as we may end up needing to make dummy classes just to get our schema looking right, but there's not much we can do about it within the limits imposed by Java's annotations.

1. I noticed that in the spec, `additionalProperties` is the only property within a Schema object that's allowed to be either `true`, `false` or another Schema object. This is reflected in our model as we have both `setAdditionalPropertiesSchema` and `setAdditionalPropertiesBoolean`.
    
    With this in mind, I think that `TrueSchema` and `FalseSchema` should not extend `Schema` and should not be in the `models` package. The spec doesn't allow `true` and `false` to be used as a schema in any place other than under `additionalProperties` and making `TrueSchema` a subtype of `Schema` gives the impression that it can be used anywhere in place of a `Schema`.
    
1. I wonder if we need `TrueSchema` or `FalseSchema` at all. Would it make sense to use `Object` to represent `true` (additional properties of any type are permitted) and `Void` to represent `false` (additional properties of any type are not permitted)?

1. Given the special nature of `additionalProperties` in the OpenAPI spec and the way it's reflected in our model, would it also make sense to have a TCK to check that the model presented to a filter is correct when `additionalProperties` is set to a boolean value with an annotation?

api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
Azquelt: Thought about it a bit more and I think we should leave it as `Class<?>` like the other attributes. I've written a bit more in my comment below.

MikeEdgar:
> > I noticed that in the spec, `additionalProperties` is the only property within a Schema object that's allowed to be either `true`, `false` or another Schema object.
> 
> On a second look, it seems like this is the case for OAS 3.0, but not for OAS 3.1?

Exactly, that's what made me think it would be good to take the approach of providing `TrueSchema` and `FalseSchema` to enable a smooth transition when OAS 3.1 support is added. The caveat is that it is only supported prior to OAS 3.1 in the way used in this PR. A more general usage should probably require that implementations provide singleton/immutable instances via the `OASFactory`, if not instances defined in the spec/API directly.

Azquelt:
> Exactly, that's what made me think it would be good to take the approach of providing `TrueSchema` and `FalseSchema` to enable a smooth transition when OAS 3.1 support is added.

I can see that making sense, but I'm not sure that this is the way we'd want to represent it. Having separate classes for `true` and `false` in the model seems really odd (even if we need to have separate classes for the annotations). In the model, I'd expect maybe separate `BooleanSchema` and `ObjectSchema` interfaces with a `Schema` super-interface, or possibly something more like the [`JsonValue`][json-value] and [`JsonObject`][json-object] interfaces in JSON-P.

[json-value]: https://javadoc.io/doc/javax.json/javax.json-api/latest/javax/json/JsonValue.html
[json-object]: https://javadoc.io/doc/javax.json/javax.json-api/latest/javax/json/JsonObject.html

I also still don't like using model classes within annotations. If we later decided to change or remove `TrueSchema` and `FalseSchema`, we'd end up affecting both the annotation classes and the model classes and I don't see what we would gain by using the same class in both the model and the annotations. I'd still really prefer it if these were separate classes in the `annotations` package that didn't extend `Schema`, even if we were to later introduce similar classes in the `model` package when we add support for OpenAPI 3.1 in a future version.

I'm fairly sure OpenAPI 3.1 will cause breaking changes in our model, so I'm not too concerned about having forward-looking support for representing `true` and `false` schemas in the model now. I'd rather be free to design that API when we're looking at supporting everything in 3.1 and can make a breaking change if that seems like the best way forward.

MikeEdgar:
> I also still don't like using model classes within annotations. If we later decided to change or remove `TrueSchema` and `FalseSchema`, we'd end up affecting both the annotation classes and the model classes and I don't see what we would gain by using the same class in both the model and the annotations. I'd still really prefer it if these were separate classes in the `annotations` package that didn't extend `Schema`, even if we were to later introduce similar classes in the `model` package when we add support for OpenAPI 3.1 in a future version.

You've convinced me :) I'll update it to keep the changes local to the `annotations` package.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -1021,4 +1021,44 @@ public void testExceptionMappers(String type) {
...
+    public void testAdditionalPropertiesTrue(String type) {
+        ValidatableResponse vr = callEndpoint(type);
+
+        String responseSchema =
+                dereference(vr, "paths.'/bookings/{id}'.get.responses.'200'", "content.'application/json'.schema");
+
+        vr.body(responseSchema, notNullValue());
+
+        String airlineSchema = dereference(vr, responseSchema, "properties.returningFlight", "properties.airline");
+        vr.body(airlineSchema + ".additionalProperties", equalTo(true));
```
Azquelt, line 1049: The OpenAPI spec states that `true` is the default value for `additionalProperties`, so I think it should be valid to omit it.

Azquelt (changes requested):
Thanks for the changes. I've added a few more comments on things that I don't think are quite right, but hopefully these are easy to fix.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
```diff
@@ -111,6 +111,42 @@ public static String dereference(ValidatableResponse vr, String path) {
         }
     }
 
+    public static String dereference(ValidatableResponse vr, String... paths) {
```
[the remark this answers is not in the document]
MikeEdgar: Let me know if this doc is clear.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
```diff
@@ -111,6 +111,42 @@ public static String dereference(ValidatableResponse vr, String path) {
...
+            }
+
+            lookup.append(path);
+            lookup.append(".$ref");
+
+            String ref = response.path(lookup.toString());
+
+            if (ref != null) {
+                absolutePath.setLength(0);
+                absolutePath.append(ref.replaceFirst("^#/?", "").replace('/', '.'));
```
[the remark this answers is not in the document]
MikeEdgar: restassured is using JsonPath for this functionality (versus Json Pointer).

api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
```diff
@@ -401,6 +428,31 @@
...
+     * Provides a Java class as implementation for additional properties that may be present in instances of this
+     * schema.
+     * 
+     * <p>
+     * If no additional properties are allowed, the value of this property should be set to {@link False False.class}
+     * which will be rendered as boolean <code>false</code> in the resulting OpenAPI document.
+     * 
+     * <p>
+     * The default value {@link True True.class} indicates that additional properties are allowed for object-type
+     * schemas.
```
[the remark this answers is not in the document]
MikeEdgar: Fixed this and also added a test to verify nothing is generated for `additionalProperties`.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
MikeEdgar: Updated test to check for both valid options for `true`. However, I am on the fence about this one. Although `true` and `additionalProperties` being absent are equivalent from the standpoint of OpenAPI semantics, it might make more sense to honor the application's explicit configuration of the `true` value and require it in the output.

Azquelt: I don't really mind either way. I added the comment because it's important that the TCK doesn't test anything that isn't specified, so if there are two valid ways of doing something the TCK must allow either.

However, looking back at the javadoc for `True`, it does look like you do explicitly require that `true` is included in the model, so I think you are ok to check for `true` in the output.
> Marker class for use in additionalProperties to indicate that the corresponding schema's additionalPropertiesBoolean value is to be set to boolean true

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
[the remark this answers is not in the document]
Azquelt: Yes, that's what I meant. If you have a `ref` where one of the keys in the path includes a `/`, it'll be replaced with `~1`. When you do the conversion to JsonPath, I think you would need to change those back.

This is likely to happen if you have a reference to an operation, because a Json Pointer to the operation will include one of the path elements which will likely include a `/`.

For example:
```openapi: 3.0.3
info: 
  title: "test"
  version: "1.0"
paths:
  '/foo/bar':
    get:
      description: "test operation"
      responses:
        200:
          description: OK
```
The test operation would be referenced as `#/paths/~1foo~1bar/get`. I think as JsonPath this would need to become `paths./foo/bar.get`.

I think paths are the only keys likely to contain slashes and I think operations are the only thing you can reference that would include a path in the reference string, as long as you're not doing that it should be fine but I do think a mention in the javadoc that it doesn't handle references with escaped characters would be helpful.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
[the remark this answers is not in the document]
Azquelt: Looks great.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
MikeEdgar: This check has been restored.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AppTestBase.java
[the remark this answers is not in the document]
MikeEdgar: Makes sense. I've updated the docs to mention that references with escapes are not supported.

MikeEdgar:
@Azquelt I think I've addressed all the comments. This has been squashed/rebased as well.

Azquelt (approved):
Looks good, thank you :smile:

PR #513, opened 2022-02-27
Support extensions as nested annotations for extensible models

### LINKED ISSUE #339: Clarify Extensible name (with/without x-)
The javadoc of `Extensible.addExtension` states:

    @param name the key used to access the extension object. Always prefixed by "x-".

This can be understood in two ways: 
1. The name passed is plain and the extensible object will prefix it with `x-` in its map.
2. The name should start with `x-`. Does this mean other should be ignored?

In the TCK examples the annotations use names with the `x-` prefix already in the name. I'd think that from the user perspective these should be plain names without the `x-` as it is more readable and avoids using illegal names by design.

### LINKED ISSUE #387: Extensions via Annotations
Annotations for models that are `Extensible` do not provide a way to list the extensions. Opening this issue for discussion in case this was not by design (I didn't see any issues about it). 

I would expect to see something like `Extension[] extensions() default {}` in each of the annotations where the equivalent model is `Extensible`.

MikeEdgar:
Fixes #339 
Fixes #387

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
@@ -108,9 +108,10 @@ public @interface Components {
     Callback[] callbacks() default {};
 
     /**
-     * The list of optional extensions.
+     * List of extensions to be added to the {@link org.eclipse.microprofile.openapi.models.Components Components} model
+     * corresponding to the containing annotation.
      *
-     * @return an optional array of extensions
+     * @return array of extensions
      * 
      * @since 3.1
      */
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
@@ -22,6 +22,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.eclipse.microprofile.openapi.annotations.callbacks.Callback;
+import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
 import org.eclipse.microprofile.openapi.annotations.headers.Header;
 import org.eclipse.microprofile.openapi.annotations.links.Link;
 import org.eclipse.microprofile.openapi.annotations.media.ExampleObject;
@@ -105,4 +106,14 @@ public @interface Components {
      * @return the reusable Callback objects.
      */
     Callback[] callbacks() default {};
+
+    /**
+     * The list of optional extensions.
+     *
+     * @return an optional array of extensions
+     * 
+     * @since 3.1
+     */
+    Extension[] extensions() default {};
+
 }
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/extensions/Extension.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/extensions/Extension.java
@@ -24,7 +24,51 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * A named extension that should be added to the OpenAPI definition.
+ * A named extension that should be added to the OpenAPI definition. The names of all extensions MUST begin with
+ * {@code x-} or else an invalid document will potentially be created.
+ * 
+ * <p>
+ * Although this annotation may currently be placed directly on a Java language element target, application developers
+ * should instead utilize the {@code extensions} property of the particular annotation that corresponds to the model
+ * being extended. Use of the annotation directly on a Java element is often ambiguous and it may result in the
+ * extension being added to an incorrect location in the OpenAPI model. Future releases of MicroProfile OpenAPI may
+ * remove the capability of placing this annotation directly on a Java element.
+ * 
+ * <p>
+ * When {@literal @}Extension annotations are used both directly on a Java element as well as within another annotation
+ * that targets the same Java element, implementations will apply only the nested extensions to the resulting model.
+ * 
+ * <p>
+ * Example of <em>preferred</em> use with {@literal @}Extension nested within an {@literal @}Schema annotation:
+ * 
+ * <pre>
+ * class MyPojo {
+ *
+ *     {@literal @}Schema(
+ *         type = SchemaType.STRING,
+ *         extensions = {@literal @}Extension(
+ *             name = "x-custom-property",
+ *             value = "custom-value")
+ *     String property1;
+ * 
+ * }
+ * </pre>
+ * 
+ * <p>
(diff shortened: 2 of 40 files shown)


MikeEdgar:
We may want to add some additional TCK tests for some random selection of annotations.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
```diff
@@ -303,4 +305,19 @@ public void testAPIResponseSchemaDefaultResponseCode(String type) {
...
+        ValidatableResponse vr = callEndpoint(type);
+
+        String opPath = "paths.'/store/inventory'.get";
+
+        vr.body(opPath, hasEntry(equalTo("x-operation-ext"), equalTo("test-operation-ext")));
+
+        vr.body(opPath + ".responses.'200'", not(hasProperty("'x-operation-ext")));
+        vr.body(opPath + ".responses.'200'", hasEntry(equalTo("x-response-ext"), equalTo("test-response-ext")));
+        vr.body(opPath + ".responses.'500'", not(hasProperty("'x-operation-ext")));
+        vr.body(opPath + ".responses.'503'", hasEntry(equalTo("x-operation-ext"), equalTo("test-operation-ext")));
```
Azquelt, line 321: Should these checks use `hasKey` rather than `hasProperty`?

MikeEdgar: I'll try that, it looks better. The test did pass on my branch in smallrye-open-api with these changes.

Azquelt:
Should we add the `extensions` array to the `@Content` annotation as well? As far as I can make out, it represents a key and value pair from the a `content` map. The value corresponds to a `MediaType` model object which is extensible.

MikeEdgar:
> Should we add the `extensions` array to the `@Content` annotation as well?

Yes, thanks for catching it.

Azquelt:
Should we add `extensions` to `@APIResponses`? I think that's allowed even though it's a repeatable container for `@APIResponse` as long as it has a default value and the `APIResponses` model object is extensible.

Azquelt:
I'm unsure whether we should add it to `@APISchemaResponse` or not. I think that corresponds to a `Response` but it's also already a shorthand and users who want to add extensions could use the full `@APIResponse` instead.

MikeEdgar:
> it's also already a shorthand and users who want to add extensions could use the full `@APIResponse` instead

That's my thinking as well.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
```diff
@@ -105,4 +106,14 @@
      * @return the reusable Callback objects.
      */
     Callback[] callbacks() default {};
+
+    /**
+     * The list of optional extensions.
+     *
+     * @return an optional array of extensions
```
Azquelt, line 113: I'm not sure if referring to extensions as "optional extensions" in the javadoc is a good idea. It's not clear in what way they're "optional".

In particular, it's not optional for implementers to include them in the model.

It is valid for the user not to use any extensions, so they're optional in that way, but that's true of many annotation parameters so I don't think it warrants calling it out here.

MikeEdgar: I'll update this, thanks. It's originally from the text of `CallbackOperation#extensions` which already exists before this PR, but I think "optional" is not the best choice to describe this. (I'll update the `CallbackOperation` JavaDoc as well).

api/src/main/java/org/eclipse/microprofile/openapi/annotations/extensions/Extension.java
```diff
@@ -24,17 +24,62 @@
...
+ * 
+ * <p>
+ * Although this annotation may currently be placed directly on a Java language element target, application developers
+ * should instead utilize the {@code extensions} property of the particular annotation that corresponds to the model
+ * being extended. Use of the annotation directly on a Java element is often ambiguous and it may result in the
+ * extension being added to an incorrect location in the OpenAPI model. Future releases of MicroProfile OpenAPI may
+ * remove the capability of placing this annotation directly on a Java element.
+ * 
+ * <p>
+ * When {@literal @}Extension annotations are used both directly on a Java element as well as within another annotation
```
Azquelt, line 38: This whole javadoc is really good. :)

I think usually I expect to see class and annotation names in a `{@code }` block (as you've done for the `extensions` property). Maybe that's a matter of style.

Azquelt (commented):
I've finished my review and I'd be happy to approve once the comments above are addressed.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/extensions/Extension.java
MikeEdgar: I prefer `{@code }` as well - this had been updated.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java
MikeEdgar: Done

tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
```diff
@@ -315,9 +314,14 @@ public void testExtensionPlacement(String type) {
 
         vr.body(opPath, hasEntry(equalTo("x-operation-ext"), equalTo("test-operation-ext")));
 
-        vr.body(opPath + ".responses.'200'", not(hasProperty("'x-operation-ext")));
+        vr.body(opPath + ".responses", hasEntry(equalTo("x-responses-ext"), equalTo("test-responses-ext")));
+        vr.body(opPath + ".responses.'200'", not(hasKey("'x-operation-ext")));
         vr.body(opPath + ".responses.'200'", hasEntry(equalTo("x-response-ext"), equalTo("test-response-ext")));
-        vr.body(opPath + ".responses.'500'", not(hasProperty("'x-operation-ext")));
+        vr.body(opPath + ".responses.'500'", not(hasKey("'x-operation-ext")));
```
Azquelt, line 320: Lines 318 and 320 have an extra `'` character (`'x-operation-ext` instead of `x-operation-ext`).

Since they're negative tests, I'd expect them to pass anyway, but I don't think they're testing what they're meant to be.

Azquelt (changes requested):
Changes look good but spotted one issue in the test.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
MikeEdgar: Good catch - fixed them.

Azquelt (approved):

PR #517, opened 2022-03-08
Test default `required` value for each parameter type

### LINKED ISSUE #355: Define default values for required property per parameter type
I couldn't find much info about query parameter handling in the spec. This example already suggests, it could be an optional parameter and I would agree: https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.adoc#requestbody
However some implementors set them required by default, just like `@PathParam` or `@FormParam` (where it's fine).
As discussed in https://github.com/smallrye/smallrye-open-api/issues/110 one could consider defining requirements in the spec for each possible parameter type.

MikeEdgar:
Fixes #355


Emily-Jiang (approved):
LGTM

Azquelt (approved):
Looks good to me as well.

PR #519, opened 2022-03-17
Better support for security requirement objects

### LINKED ISSUE #468: @SecurityRequirementsSet not working
I'm working on switching openapi v2 to openapi3 , we'll be using microprofile 2.0 with smallrye as the plugin that will generate the OAS from the code.

Now, we have different headers that are needed as a kind of authorization. I know i can put them on each resource, that seems just a lot of repetition, so i thought it was a good idea to put in the JaxRSConfiguration file as a SecurityScheme and add them as security to the OpenApiDefinition

```
@OpenAPIDefinition(
    info = @Info(
            ...
            )
    ),
    security = {
       @SecurityRequirement(name = Header1),
       @SecurityRequirement(name = Header2)
           },
    components = @Components(
             securitySchemes = {
                    @SecurityScheme(  apiKeyName = Header1 ... ) ,
                    @SecurityScheme(  apiKeyName = Header2 ....) 
    }),
    ...
    )
```
This is working, however it generates an OR, while it should be an AND (without the - )

```
security:
  - Header1: []
  - Header2: []
```
@SecurityRequirementsSet solves this problem, although i cannot use it somewhere in the @OpenApiDefinition

Is this a bug or is this intended as is. 

Thx

### LINKED ISSUE #483: Support generation of empty SecurityRequirement via annotation
Hi,

# What do I want to achieve?
I use the smallrye-openapi implemenation of the microprofile-open-api specification to generate an openapi.json from code.
I want to achieve to following output in the openapi.json:

```
"security" : [ 
  { "ApiKeyAuth" : [ ] }, 
  { "BearerTokenAuth" : [ ] }, 
  { }
]
```

# Why do I need this?
This kind of specification is needed to achieve optional authentication as per the OpenAPI specification. See here https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#openapi-object, in the description for the security element:

> "A declaration of which security mechanisms can be used across the API. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. Individual operations can override this definition. To make security optional, an empty security requirement ({}) can be included in the array."

# What do I currently have?
I can add the following annotations in my code to get the entries for "ApiKeyAuth" and "BearerTokenAuth":

```
@SecurityScheme(securitySchemeName = "ApiKeyAuth", description = "ApiKey Authentification",
        type = SecuritySchemeType.APIKEY, apiKeyName = "apikey", in = SecuritySchemeIn.HEADER)
@SecurityScheme(securitySchemeName = "BearerTokenAuth", description = "JWT Bearer Token Authentification",
        type = SecuritySchemeType.HTTP,  scheme = "bearer", bearerFormat = "JWT")
@SecurityRequirement(name = "ApiKeyAuth")
@SecurityRequirement(name = "BearerTokenAuth")
```

# What do I need?
I also need to add the empty security requirement, but there doesn't seem to be a way to add it through annotations. 
Particularly, I cannot add a SecurityRequirement without any arguments, because the name parameter is mandatory. So this doesn't work:

```
@SecurityRequirement()
```

Am I missing something, or is this a missing feature?

# What is my current workaround?
Currently I implemented a custom OASFilter that adds the empty security requirement:

```
public class CustomOASFilter implements OASFilter {

    @Override
    public Operation filterOperation(Operation operation) {
        SecurityRequirement sec = new SecurityRequirementImpl();
        List<SecurityRequirement> secList = operation.getSecurity();
        List<SecurityRequirement> newSecList = secList != null ? new ArrayList<>(secList) : new ArrayList<>();
        newSecList.add(sec);
        operation.setSecurity(newSecList);
        return operation;
    }
}
```

Azquelt:
* Make `SecurityRequirementsSet` repeatable
* Add `securitySet` parameters to OpenAPIDefinition and CallbackOperation
* Clarify that using `@SecurityRequirement` alone is equivalent to using
`@SecurityRequirementsSet` with one element
* Add TCK for `@SecurityRequirementsSet`
* Add TCK for `securitySet` in `@OpenAPIDefinition`

Fixes #468 
Fixes #483

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/OpenAPIDefinition.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/OpenAPIDefinition.java
@@ -26,6 +26,7 @@ import java.lang.annotation.Target;
 import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
 import org.eclipse.microprofile.openapi.annotations.info.Info;
 import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirement;
+import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirementsSet;
 import org.eclipse.microprofile.openapi.annotations.servers.Server;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
 
@@ -67,10 +68,24 @@ public @interface OpenAPIDefinition {
 
     /**
      * A declaration of which security mechanisms can be used across the API.
-     *
-     * @return the array of servers used for this API
+     * <p>
+     * Adding a {@code SecurityRequirement} to this array is equivalent to adding a {@code SecurityRequirementsSet} containing
+     * a single {@code SecurityRequirement} to {@link #securitySets()}.
+     * 
+     * @return the array of security requirements for this API
      */
     SecurityRequirement[] security() default {};
+
+    /**
+     * A declaration of which security mechanisms can be used across the API.
+     * <p>
+     * All of the security requirements within any one of the sets must be satisfied to authorize a request.
+     * <p>
+     * Including an empty set within this list indicates that the other requirements are optional.
+     * 
+     * @return the array of security requirement sets for this API
+     */
+    SecurityRequirementsSet[] securitySets() default {};
 
     /**
      * Any additional external documentation for the API
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/CallbackOperation.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/CallbackOperation.java
@@ -28,6 +28,7 @@ import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
 import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
 import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirement;
+import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirementsSet;
 
 /**
  * Describes a single API callback operation.
@@ -101,12 +102,28 @@ public @interface CallbackOperation {
      * A declaration of which security mechanisms can be used for this callback operation. Only one of the security
      * requirement objects need to be satisfied to authorize a request.
      * <p>
+     * Adding a {@code SecurityRequirement} to this array is equivalent to adding a {@code SecurityRequirementsSet}
+     * containing a single {@code SecurityRequirement} to {@link #securitySets()}.
+     * <p>
      * This definition overrides any declared top-level security. To remove a top-level security declaration, an empty
      * array can be used.
      * 
      * @return the list of security mechanisms for this callback operation
      */
     SecurityRequirement[] security() default {};
+
+    /**
+     * A declaration of which security mechanisms can be used for this callback operation. All of the security
+     * requirements within any one of the sets needs needs to be satisfied to authorize a request.
+     * <p>
+     * This definition overrides any declared top-level security. To remove a top-level security declaration, an empty
+     * array can be used.
+     * <p>
+     * Including an empty set within this list indicates that the other requirements are optional.
+     * 
+     * @return the list of security mechanisms for this callback operation
+     */
+    SecurityRequirementsSet[] securitySets() default {};
 
     /**
      * List of extensions to be added to the {@link org.eclipse.microprofile.openapi.models.Operation Operation} model
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
@@ -26,6 +26,9 @@ import java.lang.annotation.Target;
 
 /**
  * Specifies a security requirement for an operation.
+ * <p>
+ * Applying this annotation to a method or class is equivalent to applying a {@link SecurityRequirementsSet} annotation
+ * containing only this annotation.
  * 
  * @see <a href=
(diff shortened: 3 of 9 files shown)


MikeEdgar:
@Azquelt thanks for the PR. I'm planning to look at this together with your smallrye-open-api PR in the next few days. I think this also fixes #483, correct?

Azquelt:
Yes, I think it does. I'll note that.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/CallbackOperation.java
```diff
@@ -101,13 +102,29 @@
      * A declaration of which security mechanisms can be used for this callback operation. Only one of the security
      * requirement objects need to be satisfied to authorize a request.
      * <p>
+     * Adding an entry to this array is equivalent to adding a {@code SecurityRequirementsSet} containing only the entry
+     * to {@link #securitySet()}.
```
MikeEdgar, line 106: Same comment as above.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
```diff
@@ -26,6 +26,9 @@
 
 /**
  * Specifies a security requirement for an operation.
+ * <p>
+ * Applying this annotation to a method or class is equivalent to applying a {@link SecurityRequirementsSet} annotation
+ * containing this annotation.
```
MikeEdgar, line 31:
```suggestion
 * Applying this annotation to a method or class is equivalent to applying a {@link SecurityRequirementsSet} annotation
 * containing only this annotation.
```

api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirementsSets.java
```diff
@@ -0,0 +1,53 @@
...
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Represents an array of security requirement sets that can be specified for the operation or at definition level. Only
+ * one of requirement sets needs be satisfied.
```
MikeEdgar, line 28: > or at definition level

I know this phrase is in the equivalent `SecurityRequirements` annotation docs as well, but reading it makes it seem that this annotation is supported alongside `@OpenAPIDefinition` on an `Application` or `package-info`. What do you think?

MikeEdgar (commented):
A few comments/questions.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/OpenAPIDefinition.java
```diff
@@ -67,11 +68,25 @@
 
     /**
      * A declaration of which security mechanisms can be used across the API.
+     * <p>
+     * Adding an entry to this array is equivalent to adding a {@code SecurityRequirementsSet} containing only the entry
+     * to {@link #securitySet()}.
```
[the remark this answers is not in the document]
Azquelt: It's trying to say that adding `@SecurityRequirement(name = "foo")` to `security` is the same as adding `@SecurityRequirementSet([@SecurityRequirement(name = "foo")])` to `securitySet`.

Is it clearer like this? :
```
Adding a {@code SecurityRequirement} to this array is equivalent to adding a {@code SecurityRequirementsSet} containing a single {@code SecurityRequirement} to {@link #securitySet()}.
```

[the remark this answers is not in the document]
MikeEdgar: Yes, I think that's clearer.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirementsSets.java
Azquelt: I think it's very imprecise about where it's supported.

Would it be better to list the places where it is supported? (which would be on a resource method, on a resource class and within `@OpenAPIDefinition#securitySet`)

MikeEdgar: I think that makes sense.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/OpenAPIDefinition.java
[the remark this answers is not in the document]
Azquelt: Done.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
Azquelt: Done.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/OpenAPIDefinition.java
```diff
@@ -67,11 +68,25 @@
...
+     * A declaration of which security mechanisms can be used across the API.
+     * <p>
+     * All of the security requirements within any one of the sets must be satisfied to authorize a request.
+     * <p>
+     * Including an empty set within this list indicates that the other requirements are optional.
+     * 
+     * @return the array of security requirement sets for this API
+     */
+    SecurityRequirementsSet[] securitySet() default {};
+
```
MikeEdgar, line 89: Looking at it again, I think it might be more natural for this and the property in `CallbackOperation` to be plural, `securitySets`.

MikeEdgar (commented):
Only one thing came to mind this pass through.

MikeEdgar (approved):
Looks good, thank you!

Azquelt:
Review commits squashed

PR #523, opened 2022-04-11
Don't test format parameter

### LINKED ISSUE #503: Support for query parameter `format` is optional according to spec but is required by the TCK
The spec states that support is "suggested but optional" for the `format` query parameter on requests to the `/openapi` endpoint.

But the `PetStoreAppTest#testJsonResponseTypeWithQueryParameter` test requires that support.

Azquelt:
Remove PetStoreAppTest.testJsonResponseTypeWithQueryParameter because
support for the format parameter is optional.

Fixes #503


MikeEdgar (approved):
LGTM

PR #524, opened 2022-04-11
Allow APIResponse on resource class

### LINKED ISSUE #417: APIResponse and APIResponses should target TYPE as well

Azquelt:
`APIResponse` annotations on the method override `APIResponse` annotations
on the resource class with the same response code.

Fixes #417

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponse.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/responses/APIResponse.java
@@ -46,6 +46,11 @@ import org.eclipse.microprofile.openapi.annotations.media.Content;
  * }
  * </pre>
  * <p>
+ * When this annotation is applied to a JAX-RS resource class, the response is added to the responses defined in all
+ * OpenAPI operations which correspond to a method on that class. If an operation already has a response with the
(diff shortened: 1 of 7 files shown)


Azquelt:
Reading over this again, should we allow `@APIResponse` to be applied to an `ExceptionMapper` class as well as the `toResponse` method?

Adding it to the class seems more natural, I assume it was only applied to the method because previously the annotation only targeted methods.

MikeEdgar:
> Reading over this again, should we allow @APIResponse to be applied to an ExceptionMapper class as well as the toResponse method?

I think that makes sense.

Azquelt:
Updated to test for APIResponse on an ExceptionMapper class.

MikeEdgar (approved):
Looks good, thank you.

Issue #484, opened 2021-06-03
Examples/Demo


YunaBraska:
Since it's not intuitive to use, I would like to see some more examples how to provide each OpenApi element.
Currently I struggle to define the the general and method authorization with bearer and also to define a method description.
@Schema is also valid for methods but gets ignored when describing my method.
It would be great if you could provide more documentation or even improve annotations and their documentation.
Thanks

Emily-Jiang:
Maybe a wiki page should be written for the suggested items. The 3rd bulletin point will trigger a TCK to be added (#508 )

PR #525, opened 2022-04-14
Tests for hidden schemas

### LINKED ISSUE #508: TCK for @Schema(hide-true) behaviour
Some impl does to honor: a field @Schma(hide = true) . A TCK should be added to force this behaviour.

Azquelt:
* Test that a field annotated with @Schema(hidden = true) is not
  included in the schema for the enclosing class
* Test that a resource method parameter annotated with @QueryParam and
  @Schema(hidden = true) creates a parameter with no schema in the
  openapi document
* Test that a resource method parameter annotated with @Schema(hidden =
  true) creates a request body with no schema in the openapi document

I wasn't hugely happy with this issue, the spec isn't clear what "hidden" means and it's hard to interpret for all the locations that the `@Schema` annotation could be applied. The lack of structure in the TCKs is also unhelpful in deciding how to add or structure a new set of tests.

In the end I've tested `hidden = true` in the scenarios listed above and required that the `schema` parameter is not present in the openapi document which seems reasonable, but it could also be tested in the following locations:

* Applied to a class which is used as a field in a different class
* Applied to a class which is used as a request body
* Applied to a class which is used as a request parameter
* Applied to a class which is used as a response body
* Applied to a resource method (where it would affect the response body)

Fixes #508

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/model/User.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/model/User.java
@@ -48,6 +48,9 @@ public class User {
     @Schema(required = true, example = "1")
     private int status;
 
+    @Schema(hidden = true)
+    private String undocumentedProperty;
+
     /**
      * Creates a User instance with the parameters specified.
      * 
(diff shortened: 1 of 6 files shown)


MikeEdgar (approved):
LGTM. I agree that it's not simple to cover all the locations where `@Schema` could be used. The original issue #484 wasn't specific on the problem, so I think this PR is fine at this point in time.

PR #528, opened 2022-04-19
Add a recommended location for any UI

### LINKED ISSUE #334: Visualize the OpenAPI document at a standard URI
At our weekly hangout today we discussed the possibility of a spec-defined URL where a user could visualize the generated OpenAPI document.  The idea is that the spec would indicate how to enable/disable such a feature **and** specify where a user could find such a visualization.  This would probably be similar to validation in approach (config property to enable/disable and well defined endpoint to load the UI) but is also open for discussion of course.

As an example of what this might look like for a vendor - Open Liberty provides an endpoint (e.g. `/openapi/ui` that results in Swagger UI being opened with the generated OpenAPI document.

Azquelt:
Fixes #334


MikeEdgar (approved):
Looks good, thanks.

PR #529, opened 2022-04-25
Replace JAX-RS references

### LINKED ISSUE #527: Update references to JAX-RS
Since 3.0, MP OpenAPI depends on Jakarta EE 9, so should all the references to JAX-RS be changed to instead reference Jakarta RESTful Web Services?

Azquelt:
Replace spec references to JAX-RS with Jakarta RESTful Web Services.

I've used the full name for the first mention and otherwise abbreviated to RESTful Web Services.

Fixes #527


Emily-Jiang (changes requested):
I would like the name replacement for JAX-RS agreed among the MP specs and use the same names across the board. Please don't mege in the PR as yet.

PR #536, opened 2022-06-14
Add 3.1 release notes

### LINKED ISSUE #535: Create release notes for 3.1
We need to update the release notes here: https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/release_notes.asciidoc

Azquelt:
Fixes #535 

Also added labels to the subheadings in the release notes so that they'll have stable anchors. When you repeat a header like "API Changes", the first instance gets the anchor `api_changes` and the second gets `api_changes_2`. This means that the anchors will change when we add a new version to the release notes.

### DIFF
--- a/spec/src/main/asciidoc/release_notes.asciidoc
+++ b/spec/src/main/asciidoc/release_notes.asciidoc
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+// Copyright (c) 2016-2022 Contributors to the Eclipse Foundation
 //
 // See the NOTICE file(s) distributed with this work for additional
 // information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


MikeEdgar (approved):
Looks good, thank you @Azquelt

spec/src/main/asciidoc/release_notes.asciidoc
```diff
@@ -16,19 +16,48 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+[[release_notes_31]]
+== Release Notes for Microprofile OpenAPI 3.1
```
Emily-Jiang, line 20:
```suggestion
== Release Notes for MicroProfile OpenAPI 3.1
```

Emily-Jiang (approved):

spec/src/main/asciidoc/release_notes.asciidoc
Azquelt: Will fix this here and in the other headings

Emily-Jiang (approved):

PR #541, opened 2022-06-23
Limit Bean Validation processing requirement

### LINKED ISSUE #540: Does Bean Validation integration need to be optional for 3.1?
When we added Bean Validation integration in #482, we assumed that Bean Validation would be included in MicroProfile.

We're now not expecting that to be the case when we release MP OpenAPI 3.1. Do we need to make the Bean Validation integration only applicable on runtimes which support bean validation?

I imagine we would:
* Reword the spec language to indicate that integration is required if Bean Validation is present
* Add the Bean Validation integration tests to a test group so that they can be easily excluded
* Add documentation to the tck readme for disabling the Bean Validation tests for runtimes which do not support Bean Validation

Azquelt:
Only require Bean Validations to be processed for creating the OpenAPI
document if support for Jakarta Bean Validation is actually present.

* Add spec wording
* Add Bean Validation tests to a group
* Update TCK readme to explain how to exclude the group

Also fix formatting error in the bean validation section of the spec,
bullet lists need to be preceded by a blank line.

Fixes #540

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
@@ -589,7 +589,8 @@ For more samples please see the https://github.com/eclipse/microprofile-open-api
 
 In some cases, additional schema restrictions can be inferred from Jakarta Bean Validation annotations and used to enhance the generated OpenAPI document.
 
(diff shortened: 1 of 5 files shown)


Azquelt:
This will need an update if it's merged after #538

MikeEdgar (approved):
LGTM

Azquelt:
Updated to add the bean validation group to the test added in #538

PR #545, opened 2022-08-23
Make most archives not testable

### LINKED ISSUE #546: Allow the TCK to work using the new arquillian REST protocol
I came across an issue when trying to run the TCK using the new arquillian REST 3.0 protocol (which uses a jax-rs resource instead of a servlet to communicate between the client and server).

This extra rest resource is put into every test deployment which means it's in the application and gets scanned and added to the `/openapi` endpoint, which breaks some of the tests because there are unexpected entries in the openapi document.

The fix is to mark the deployment as not "testable" (which means there aren't any tests which will run _within_ the deployment) so that arquillian doesn't include the extra REST resource in the application.

This should be changed for all of our deployments apart from `ModelConstructionTest` and `OASFactoryErrorTest`

Azquelt:
A "testable" archive is one that arquillian adds extra framework code to
in order to run tests within the server runtime environment.

Most of the MP OpenAPI test classes don't include any tests which run on
the server side, they deploy an app and then query the /openapi endpoint
from the client side.

In addition, it causes problems when using the new arquillian REST 3.0
protocol. When using this protocol, an additional REST resource is
deployed within testable deployments which is scanned and added to the
openapi documentation which causes some of the tests to fail.

Fixes #546

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
@@ -43,7 +43,6 @@ import java.util.List;
 
 import org.hamcrest.collection.IsMapWithSize;
 import org.jboss.arquillian.container.test.api.Deployment;
-import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.testng.annotations.Test;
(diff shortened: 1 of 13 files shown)


MikeEdgar:
This seems to cause the SmallRye [`DeploymentProcessor`](https://github.com/smallrye/smallrye-open-api/blob/53522ec37e219aefc9b2266c44566e3259df2cd2/testsuite/tck/src/test/java/io/smallrye/openapi/tck/DeploymentProcessor.java#L55) to no longer be executed. The [`ArquillianExtension`](https://github.com/smallrye/smallrye-open-api/blob/53522ec37e219aefc9b2266c44566e3259df2cd2/testsuite/tck/src/test/java/io/smallrye/openapi/tck/ArquillianExtension.java#L8) is still called, which is where the processor is registered. Any idea why that might be the case?

Azquelt:
Yeah, I'd noticed that and have just been poking through the arquillian code. It looks like a registered `ApplicationArchiveProcessor` is only called for a testable archive.

You're using `DeploymentProcessor` to actually add your implementation into the deployed application, right? So you need to do that to all archives, not just testable ones.

I'm having a look to see if we can observe `BeforeDeploy` and update the archive there instead.

MikeEdgar:
That's right. From what I can tell, `BeforeDeploy` provides read-only access to the archive.

Azquelt:
It seems to allow me to edit it: #1224

MikeEdgar (approved):

PR #547, opened 2022-09-23
Update to use the parent pom 2.6 and prepare for the release 3.1

Emily-Jiang:


MikeEdgar:
We'll need to drop JDK 8 from the CI process: https://github.com/eclipse/microprofile-open-api/blob/master/.github/workflows/build.yml#L11

It might be good to add 17 at the same time as well.

Azquelt (approved):

Azquelt:
Looks like the parent pom is still specifying Java 8, which doesn't work when there are EE10 dependencies. https://github.com/eclipse/microprofile-parent/blob/f5dcc33183f3b1f274fc1d526a9c21f581f6fa35/pom.xml#L35-L36

MikeEdgar:
> Looks like the parent pom is still specifying Java 8, which doesn't work when there are EE10 dependencies. https://github.com/eclipse/microprofile-parent/blob/f5dcc33183f3b1f274fc1d526a9c21f581f6fa35/pom.xml#L35-L36

That should be OK. This project should still create JDK 8-compatible binaries, but the CI won't be able to run with JDK 8 (see my earlier comment). I am fine to bump the output to Java 11 bytecode, but I thought that would be avoided unless the API/features requires it.

Azquelt:
Ah, that's ok then, I'd wrongly interpreted the failures on the build from ci.eclipse.org. I guess it just needs updating to use Java 11 as well.

Emily-Jiang:
Let me remove Java 8 from CI job as the mininum Java version for Jakarta EE 10 is Java 11.

Emily-Jiang:
I have fixed the CI jobs to only run with Java 11 as the apis from CDI, JAX-RS in Jakarta EE 10 compile with Java 11.

MikeEdgar (approved):
Changes look good to me. I'll also try this PR with SmallRye OpenAPI to be aware of any problems. It looks like the Eclipse Jenkins job is also still running with JDK 8.

Emily-Jiang:
> Changes look good to me. I'll also try this PR with SmallRye OpenAPI to be aware of any problems. It looks like the Eclipse Jenkins job is also still running with JDK 8.

yep, looking at it right now.

Emily-Jiang:
I had a look at the Jenkins job and it seems to do with the microprofile parent java compilation. I have updated it to be 11 [here](https://github.com/eclipse/microprofile-parent/pull/65). @MikeEdgar if you can review it, it will be good as I can do another microprofile-parent 3.0-RC2 to test it out. thanks

MikeEdgar:
@Emily-Jiang , I updated the Jenkins job to use Eclipse Temurin JDK 11. The PR might need another push or open/close to trigger another build.

MikeEdgar:
Re-reading the thread `Version changes for new Jakarta EE versions`, I'm now questioning whether this update should happen. Without changes to the API/spec in this project, I think compiling against the old Jakarta dependency is the correct thing to do and will not prevent usage with newer Jakarta versions.

Emily-Jiang:
> Re-reading the thread `Version changes for new Jakarta EE versions`, I'm now questioning whether this update should happen. Without changes to the API/spec in this project, I think compiling against the old Jakarta dependency is the correct thing to do and will not prevent usage with newer Jakarta versions.

@MikeEdgar The [release plan](https://docs.google.com/spreadsheets/d/1yYI5pdRCoOT5GaZ6KMNTfg3mQ5Hwxxv45SW5WBsZAmY/edit#gid=1995986219) for MP 6.0 has specified that all updated specs need to align with Jakarta EE 10 core profile while the unupdated specs don't need to such as Config, Rest Client, Health and Fault Tolerance. In this case, MP Open API 3.1 must align with Jakarta EE 10 core profile and will need to use the microprofile-parent 3.0 release. I have explained the reason why the 3.0 parent pom needs to sync up with Jakarta EE 10 and moving Java version to 11.

Emily-Jiang:
@Azquelt @MikeEdgar I have updated the PR accordingly using the parent pom of 2.6

MikeEdgar (approved):
Looks good. Thanks @Emily-Jiang

Issue #557, opened 2023-02-08
Question: Ambiguous Resource Methods in TCK, airlines UserResource


aubi:
In [UserResource](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/UserResource.java#L250), two methods have the same path (e.g. refusing to deploy):

```
    @GET
    @Path("/{username}")
...
    public Response getUserByName(
```
and
```
    @GET
    @Path("/{id}")
...
    public Response getUserById(
```
The only difference is, that `getUserById` defines the parameter as `int`, while the first as `String`.

**Is the path parameter type enough to distinguish the methods?** So if the parameter is convertible to int, use `getUserById`; use `getUserByName` otherwise? To me it seems wrong, because `getUserByName` will never work for username made from digits.

Azquelt:
Huh, I hadn't noticed this and I'm not sure if it's valid - would need to check the Jakarta RESTful WS spec for that - but it at least seems unnecessary to test something so confusing here.

Before changing it or removing one of the methods, we'd have to go through and check if any tests are expecting these entries to be present and update or remove those tests.

PR #559, opened 2023-03-17
distinguish resources paths in airlines app, by username and id

aubi:
In the original version, the testing apps cannot deploy to an application server (Payara in my case), as the paths '/user/{username}' and '/user/{id}' are ambiguous.

According to the reaction in the issue https://github.com/eclipse/microprofile-open-api/issues/557, I tried to suggest a fix.


MikeEdgar (approved):
Looks good to me. @Azquelt , any concerns?

Azquelt:
Sorry for the slow reply, I was on vacation.

No concerns from me. The changes look good and I ran the TCK with this change against OpenLiberty and it still passed.

PR #560, opened 2023-04-13
TCK deps: MP Rest Client API to 3.0.1, JAXB to 3.0.1, BV to 3.0.2

### LINKED ISSUE #551: Parent POM MP Rest Client version property is not in use and outdated
There is a defined property for the MP REST Client version here:

https://github.com/eclipse/microprofile-open-api/blob/8c8adaad12a7dd4cdd0bd12bb3967e60d62db98f/pom.xml#L39

But it is not in use here in the TCK:
https://github.com/eclipse/microprofile-open-api/blob/8c8adaad12a7dd4cdd0bd12bb3967e60d62db98f/tck/pom.xml#L53-L58

The used version is also outdated, as in MP 6.0 it is planed to use 3.0.1 (instead of 3.0), that is not released yet. So this use of combining component spec service releases in the platform release without following the dependency tree and defining a minimum version (aka using a range in fact) can create issues.

Regarding the property use, this could be fixed by using it in the TCK (then it may be renamed too, for better recognition) or it could be deleted simply (other TCK dependencies have no parent POM version property too, like Jakarta Validation).

The referenced version should be in sync with the version that will be defined in the MP release, so 3.0.1 here and that means, releasing the MP (Platform) requires to release it's compontents from the bottom to the top of the dependency tree to keep them in sync.
As this affects the TCK here only, there is a risk of testing with or against an old version, that could create an issue late in the MP release process, when the MP implementation tests combined specs with other (newer) versions (or even worse: testing against the same old versions, but delivering newer ones).
In short: The (test) environment for the component spec should be the same as the platform environment.

MikeEdgar:
Fixes #551

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,12 +31,14 @@
     <properties>
         <inceptionYear>2017</inceptionYear>
         <osgi-annotation.version>1.1.1</osgi-annotation.version>
-        <!-- Versions of TCK dependencies -->  
+        <!-- Versions of TCK dependencies -->
         <rest-assured.version>4.3.0</rest-assured.version>
         <java-hamcrest.version>2.0.0.0</java-hamcrest.version>
         <httpclient.version>4.5.2</httpclient.version>
         <jackson.version>2.10.1</jackson.version>
-        <restclient.version>3.0</restclient.version>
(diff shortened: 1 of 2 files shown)


MikeEdgar:
Build please

Azquelt (approved):

PR #561, opened 2023-04-13
Add missing `yml` suffix to conditions for serving `/openapi` endpoint

### LINKED ISSUE #553: Deviation in openapi file postfix definition
There are two sections that define the openapi file postfix (.yaml, .yml,.json), but they are not consistent:

https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc#location-and-formats

In sentence "Vendors are required to fetch a single document named openapi with an extension of yml, yaml or json, inside the application module’s root META-INF folder.".

and

https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc#overview

In list element "one of the allowed static files is present, i.e. META-INF/openapi.(json|yaml)".

The last should be changed to "one of the allowed static files is present, i.e. META-INF/openapi.(json|yaml|yml)" or in the first occurence the "yml" should be removed instead (but that might be a breaking change potentially).

MikeEdgar:
Fixes #553

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
@@ -761,7 +761,7 @@ following conditions are met:
 
 - an `OASModelReader` has been configured with `mp.openapi.model.reader`
 - an `OASFilter` has been configured with `mp.openapi.filter`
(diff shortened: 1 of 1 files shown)


MikeEdgar:
Build please

Azquelt (approved):

PR #562, opened 2023-04-18
TCK: fix several incorrectly used annotations and associated tests

### LINKED ISSUE #322: Clean up the Pet Store (and possibly Airlines) app annotations
The Pet Store app (possibly also the Airlines app, I haven't checked) has a number of `@Parameter` annotations that are likely incorrect because the app was taken from the old Swagger 2.0 codebase.  For example, there are a number of `@Parameter` annotations on method arguments which would typically be represented as a request body.  In Swagger 2.0 that was correct, because request bodies were defined as "body" style parameters.  In 3.0, however, there is an actual request body.  There may be other annotations that do not make sense in 3.0 - so we should review the code.

MikeEdgar:
Closes #322 

There are several remaining issues reported by the OpenAPI validator [1] including:
- `/user/special` missing schemas, this is intentional to test hidden schema functionality
- `/user` POST request body encoding includes an extension `x-encoding` flagged invalid by the parser. This appears to be a bug [2]

[1] https://validator.swagger.io/
[2] https://github.com/OAI/OpenAPI-Specification/issues/3251

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
@@ -161,10 +161,10 @@ public class ReviewResource {
     @Produces("application/json")
     public Response getReviewByUser(
             @Parameter(name = "user", description = "username of the user for the reviews", in = ParameterIn.PATH,
-                       content = @Content(examples = @ExampleObject(name = "example", value = "bsmith")), examples = {
+                       content = @Content(examples = {
                                @ExampleObject(name = "example1", value = "bsmith"),
                                @ExampleObject(name = "example2",
-                                              value = "pat@example.com")}) @PathParam("user") String user,
+                                              value = "pat@example.com")})) @PathParam("user") String user,
             @QueryParam("minRating") Integer minRating,
             @HeaderParam("If-Match") String ifMatch,
             @CookieParam("trackme") String trackme) {
@@ -201,8 +201,7 @@ public class ReviewResource {
     @Operation(operationId = "getReviewByAirline", summary = "Get all reviews by airlines")
     @Parameter(name = "airline", description = "name of the airlines for the reviews", required = true,
                in = ParameterIn.PATH,
-               content = @Content(examples = @ExampleObject(name = "example", value = "Acme Air")),
-               example = "Acme Air")
+               content = @Content(examples = @ExampleObject(name = "example", value = "Acme Air")))
     @APIResponse(responseCode = "200", description = "Review(s) retrieved",
                  content = @Content(schema = @Schema(implementation = Review.class)))
     @APIResponse(responseCode = "404", description = "Review(s) not found")
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/UserResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/UserResource.java
@@ -152,10 +152,7 @@ public class UserResource {
                          content = @Content(mediaType = "application/json",
                                             schema = @Schema(type = SchemaType.ARRAY, implementation = User.class,
                                                              nullable = true, writeOnly = true, minItems = 2,
-                                                             maxItems = 20, uniqueItems = true),
-                                            encoding = @Encoding(name = "firstName", contentType = "text/plain",
-                                                                 style = "form", allowReserved = true,
-                                                                 explode = true))) User[] users) {
+                                                             maxItems = 20, uniqueItems = true))) User[] users) {
         for (User user : users) {
             userData.addUser(user);
         }
@@ -279,7 +276,7 @@ public class UserResource {
                                parameters = @LinkParameter(name = "userId", expression = "$request.path.id"),
                                extensions = @Extension(name = "x-link", value = "test-link")),
                          @Link(name = "Review", description = "The reviews provided by user",
-                               operationRef = "/db/reviews/{userName}",
+                               operationRef = "#/paths/~1reviews~1users~1{user}/get",
                                parameters = @LinkParameter(name = "path.userName",
                                                            expression = "$response.body#userName"),
                                requestBody = "$request.path.id", server = @Server(url = "http://example.com"))
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
@@ -19,6 +19,7 @@ import org.eclipse.microprofile.openapi.annotations.ExternalDocumentation;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.callbacks.Callback;
 import org.eclipse.microprofile.openapi.annotations.callbacks.CallbackOperation;
+import org.eclipse.microprofile.openapi.annotations.enums.ParameterIn;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.enums.SecuritySchemeIn;
 import org.eclipse.microprofile.openapi.annotations.enums.SecuritySchemeType;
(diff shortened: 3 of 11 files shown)


MikeEdgar:
@Emily-Jiang , let me know if you see anything changing here that would be an issue for a micro-release, v3.1.1

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
```diff
@@ -161,10 +161,10 @@ public Response getReviewById(
     @Produces("application/json")
     public Response getReviewByUser(
             @Parameter(name = "user", description = "username of the user for the reviews", in = ParameterIn.PATH,
-                       content = @Content(examples = @ExampleObject(name = "example", value = "bsmith")), examples = {
+                       content = @Content(examples = {
                                @ExampleObject(name = "example1", value = "bsmith"),
                                @ExampleObject(name = "example2",
-                                              value = "pat@example.com")}) @PathParam("user") String user,
+                                              value = "pat@example.com")})) @PathParam("user") String user,
```
Azquelt, line 167: This slightly expands the test as we're now testing for multiple examples within the `@Content`.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/UserResource.java
```diff
@@ -279,7 +276,7 @@ public Response getUserByName(
                                parameters = @LinkParameter(name = "userId", expression = "$request.path.id"),
                                extensions = @Extension(name = "x-link", value = "test-link")),
                          @Link(name = "Review", description = "The reviews provided by user",
-                               operationRef = "/db/reviews/{userName}",
+                               operationRef = "#/paths/~1reviews~1users~1{user}/get",
                                parameters = @LinkParameter(name = "path.userName",
                                                            expression = "$response.body#userName"),
```
Azquelt, line 281:
```suggestion
                               operationRef = "#/paths/~1reviews~1users~1{user}/get",
                               parameters = @LinkParameter(name = "path.user",
                                                           expression = "$response.body#userName"),
```
Since the path parameter is named `user`.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
```diff
@@ -150,7 +153,7 @@ public void write(OutputStream output) throws IOException {
     public Response deletePet(
             @Parameter(name = "apiKey", description = "authentication key to access this method",
                        schema = @Schema(type = SchemaType.STRING, implementation = String.class, maxLength = 256,
-                                        minLength = 32)) @HeaderParam("api_key") String apiKey,
+                                        minLength = 32)) @HeaderParam("apiKey") String apiKey,
```
Azquelt, line 156: It's a bit weird that we have `name` set on the `@Parameter` annotation at all, given that the name can be found from the `@HeaderParam`, but I guess if the test does have it set, the two should match.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
```diff
@@ -170,7 +173,7 @@ public Response deletePet(
                                     schema = @Schema(implementation = ApiResponse.class)))
     @RequestBody(name = "pet",
                  content = @Content(mediaType = "application/json", schema = @Schema(implementation = Pet.class),
-                                    examples = @ExampleObject(ref = "http://example.org/petapi-examples/openapi.json#/components/examples/pet-example")),
+                                    examples = @ExampleObject(ref = "#/components/examples/pet-example")),
```
Azquelt, line 176: Using an external ref like this should be valid, right?

The resulting document won't validate because it points to an invalid URL, but this looks like the only test we have that an external reference from an annotation is passed through to the output.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
```diff
@@ -191,33 +194,37 @@ public Response addPet(Pet pet) {
     })
     @Operation(summary = "Update an existing pet", description = "Update an existing pet with the given new attributes")
     public Response updatePet(
-            @Parameter(name = "petAttribute", description = "Attribute to update existing pet record", required = true,
-                       schema = @Schema(implementation = Pet.class)) Pet pet) {
+            @RequestBody(name = "petAttribute", description = "Attribute to update existing pet record",
+                         required = true,
+                         content = @Content(schema = @Schema(implementation = Pet.class))) Pet pet) {
```
Azquelt, line 199: I don't think `name` should be set when using `@RequestBody` on a parameter like this.

Also, although the original annotation is definitely wrong, replacing it with the correct annotation is a little close to expanding the test. As nothing appears to assert this, I'd be tempted to just remove the offending annotation.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
```diff
@@ -191,33 +194,37 @@ public Response addPet(Pet pet) {
...
+                                                @ExampleObject(name = "Available", value = "available",
+                                                               summary = "Retrieves all the pets that are available"),
+                                                @ExampleObject(name = "Pending", value = "pending",
+                                                               summary = "Retrieves all the pets that are pending to be sold"),
+                                                @ExampleObject(name = "Sold", value = "sold",
+                                                               summary = "Retrieves all the pets that are sold")
+                                        })
                        }, allowEmptyValue = true) @Extension(name = "x-mp-parm1", value = "true") String status) {
         return Response.ok(petData.findPetByStatus(status)).build();
     }
```
Azquelt, line 230: This whole method looks fishy. There's no JAX-RS param annotation on the `status` parameter, so it should be the request body, which for a `GET` request is questionable at best.

I think an implementation could reasonably throw an exception if a `@Parameter(in=Query)` annotation is attached to a request body parameter, so I don't think this is valid.

I can't see any test that asserts anything about this method so I would advocate to remove it in a micro release instead of trying to fix it.

Azquelt (commented):
Thanks for doing this and picking out the invalid parts of the TCK app.

I have a few concerns where it looks like we could be adding a test rather than fixing one which we wouldn't want to do in a micro release and a few where I'm unsure if the change is correct.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
MikeEdgar: > I can't see any test that asserts anything about this method so I would advocate to remove the whole method in a micro release instead of trying to fix it.

I agree here as well. I was trying to minimize changes, but if it's fine to do it in a micro release, taking this method out would be better.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
MikeEdgar: Restored back to the original external ref.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
MikeEdgar: Removed the `name`

MikeEdgar: > replacing it with the correct annotation is a little close to expanding the test. As nothing appears to assert this, I'd be tempted to just remove the offending annotation.

Since there aren't any existing assertions on the body for this operation, I think we should keep it for now. Although it's likely to alter the output of an implementation, it's more in the spirit of this PR. It would probably be a good exercise to go through a second time and prune out methods/annotations that are not actually part of the tests.

Let me know if that sounds OK to you.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
MikeEdgar: Removed the method.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java
MikeEdgar: Restore to single example with the parameter examples removed.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetResource.java
Azquelt: ok

Azquelt (approved):
Looks good to me, thanks.

tjquinno (approved):
LGTM, too.

Issue #552, opened 2022-11-17
JAX-RS version in spec document is outdated


JanWesterkamp-iJUG:
The version number 2.0 of JAX-RS is outdated:

https://github.com/eclipse/microprofile-open-api/blob/master/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc#documentation-mechanisms

In sentence: "The MP OpenAPI specification requires vendors to produce a valid OpenAPI document from pure JAX-RS 2.0 applications."

To prevent that happend again it should be referenced from a maven dependency version directly into the spec document via asciidoc attribute or in this case here the version could be deleted simply.

PR #563, opened 2023-04-20
Use full spec name once for JAX-RS

### LINKED ISSUE #527: Update references to JAX-RS

Azquelt:
The full spec name is Jakarta RESTful Web Services. We should use the
full name at least once and then abbreviate it to JAX-RS (which is the
name everyone knows).

Also remove a reference specifically to JAX-RS 2.0.

Fixes #527


MikeEdgar (approved):
Looks good to me, also closes #552

PR #566, opened 2023-06-30
Bump microprofile-parent to 2.7, remove Java 8 GitHub CI

### LINKED ISSUE #565: Bump microprofile-parent version and update GitHub CI to align
The MP release process uses a newer version of Maven, which requires an update to the impsort-maven-plugin [1]. The plugin update can be obtained by updating to the latest 2.x release of the microprofile-parent. Further, the current impsort plugin version is no longer compatible with Java 8. As a result, this project will drop JDK 8 from the GitHub CI, but will continue to generate Java 8 bytecode until it is decided that migration to Java 11 output is necessary.

This issue tracks the change to (1) bump the microprofile-parent version and (2) drop Java 8 from the GitHub CI workflow.

[1] https://github.com/revelc/impsort-maven-plugin/issues/64

MikeEdgar:
Closes #565


MikeEdgar:
Note large number of whitespace differences due to formatter update.

Azquelt (approved):

PR #570, opened 2023-08-10
Extensions on methods don't cascade to responses

### LINKED ISSUE #556: Extensions propagation in APIResponse(s)
TCK tests requires, that annotation `Extension` is propagated down to `APIResponses.APIResponse.extensions`, but there is no reason for this.

TCK contains this piece of code in [`tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetStoreResource.java`](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/PetStoreResource.java#L65):

```
    @GET
    @Path("/inventory")
    @Produces({"application/json", "application/xml"})
    @APIResponses(extensions = @Extension(name = "x-responses-ext", value = "test-responses-ext"), value = {
            @APIResponse(responseCode = "200", description = "successful operation",
                         extensions = @Extension(name = "x-response-ext", value = "test-response-ext")),
            @APIResponse(responseCode = "500", description = "server error", extensions = {}),
            @APIResponse(responseCode = "503", description = "service not available",
                         content = @Content(extensions = @Extension(name = "x-notavailable-ext", value = "true"))),
    })
    @Operation(summary = "Returns pet inventories by status",
               description = "Returns a map of status codes to quantities")
    @Extension(name = "x-operation-ext", value = "test-operation-ext")
    public java.util.Map<String, Integer> getInventory() {
```

One of the tests checks, if the annotation `@Extension(name = "x-operation-ext", value = "test-operation-ext")` was propagated down to the "503" response in [PetStoreAppTest.java](https://github.com/eclipse/microprofile-open-api/blob/master/tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java#L294):

```
vr.body(opPath + ".responses.'503'", hasEntry(equalTo(X_OPERATION_EXT), equalTo(TEST_OPERATION_EXT)));
```

This seems strange to me -- why `@Extension` on method level should be propagated to `@APIResponses`?

I can imagine, that `x-responses-ext`, defined on `APIResponses` is propagated to all responses, which don't specify otherwise (only 503 in this case), but in fact I didn't see any support for this behavior in the [OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md); the javadoc warns, that this usage can lead to wrong behavior, but IMHO doesn't say anything about propagation.

Azquelt:
There's nothing in the spec to specify that an `@Extension` annotation placed on a method should result in that extension key and value being added to the responses of the corresponding operation.

Remove the test that was asserting this behaviour.

Fixes #556

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/PetStoreAppTest.java
@@ -291,7 +291,6 @@ public class PetStoreAppTest extends AppTestBase {
         vr.body(opPath + ".responses.'200'", not(hasKey(X_OPERATION_EXT)));
         vr.body(opPath + ".responses.'200'", hasEntry(equalTo("x-response-ext"), equalTo("test-response-ext")));
         vr.body(opPath + ".responses.'500'", not(hasKey(X_OPERATION_EXT)));
-        vr.body(opPath + ".responses.'503'", hasEntry(equalTo(X_OPERATION_EXT), equalTo(TEST_OPERATION_EXT)));
         vr.body(opPath + ".responses.'503'.content.'application/json'",
                 hasEntry(equalTo("x-notavailable-ext"), equalTo("true")));
         vr.body(opPath + ".responses.'503'.content.'application/xml'",


MikeEdgar (approved):
LGTM, thank you @Azquelt

Issue #574, opened 2023-11-27
Update JAX-RS to Jakarta RESTful Web Services with Jakarta EE 10 alignment


Emily-Jiang:
When this spec consumes the MP-Parent 3.x, it will align with Jakarta EE 10 Core Profile. With this, the reference to JAX-RS will need to be updated to Jakarta RESTful Web Services. The first occurrence will be Jakarta RESTful Web Services (Jakarta REST). The future occurrence will be Jakarta REST.

PR #577, opened 2024-01-09
Update parent, add module-info, streamline OSGi annotation dependency

### LINKED ISSUE #573: MP Parent version deviates from MP Parent TCK BOM version
The MP Parent was updated from 2.6 to 2.7, but the tck-bom version is untouched at 2.6.
These versions should be in sync.

Meanwhile, both versions are outdated, the current 2.x version is 2.8.

In gerneral, when updating the MP Parent version, the propery https://github.com/eclipse/microprofile-open-api/blob/0649b7f6d299a33c58fef56e19e17786227a0560/pom.xml#L43[version.microprofile.tck.bom] should be updated to the same value.
BTW, the property name deviates from Maven naming conventions, so renaming this property to **microprofile.tck.bom.version** would make sense too.

MikeEdgar:
Update parent, add module-info, streamline OSGi annotation dependency. Removes dependency on the `biz.aQute.bnd` artifact as the `osgi.annotation` should be the only one necessary.

Fixes #573 

I will also submit a follow-up PR with the changes for #574 once this is merged.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -28,10 +28,26 @@
     <description>MicroProfile OpenAPI API :: API</description>
 
     <dependencies>
-        <dependency> 
-            <groupId>biz.aQute.bnd</groupId>
-            <artifactId>biz.aQute.bnd.annotation</artifactId>
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.annotation</artifactId>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-checkstyle-plugin</artifactId>
+                    <version>3.2.1</version>
+                    <configuration>
+                        <!-- https://github.com/checkstyle/checkstyle/issues/8240 -->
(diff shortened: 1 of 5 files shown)


Azquelt (approved):
This looks sensible to me

Azquelt:
You've left this in draft, was there anything you still wanted to change before it's merged?

MikeEdgar:
No, nothing additional to change.

Emily-Jiang:
You should use the mp-parent 2.x not 3.x as 3.x will pull in Jakarta EE 10 api but this spec does not use any Jakarta EE 10 new APIs. Therefore, it should be built with Jakarta EE 9.1.

MikeEdgar:
@Emily-Jiang are we able to set the compiler release to 11 since 2.x is still using 1.8? Otherwise we will need to remove the module-info and possibly revert the OSGi dependency changes.

Emily-Jiang:
> @Emily-Jiang are we able to set the compiler release to 11 since 2.x is still using 1.8? Otherwise we will need to remove the module-info and possibly revert the OSGi dependency changes.

yes, you can override it to be 11 in your own pom.xml

Azquelt:
The module-info appears to cause some trouble building the aggregated javadoc: eclipse/microprofile#443

I'm looking into it, but if anyone has any ideas I'd appreciate it.

PR #581, opened 2024-01-17
Deprecate `spi` module and contents

### LINKED ISSUE #576: Consider deprecating/removing the `spi` module
The `spi` module contains an out-of-date single class that is also present in the `api` module. Maybe we can:
- remove `spi`
- provide a Maven replication from `spi` to `api`

MikeEdgar:
Closes #576

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -25,7 +25,7 @@
 
     <artifactId>microprofile-openapi-api</artifactId>
     <name>MicroProfile OpenAPI API</name>
-    <description>MicroProfile OpenAPI API :: API</description>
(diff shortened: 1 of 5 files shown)


MikeEdgar:
> I wonder if there's anything we need to do to ensure that when we build the full MicroProfile Javadoc we pick up the API version of this package rather than the SPI one with the deprecation warnings.

Good question, where is it generated and hosted?

Azquelt:
Hmm, actually never mind, it doesn't look like the MicroProfile project builds a combined Javadoc anywhere.

We do build one for the Open Liberty docs, but I can poke our docs team to find out how they do it and ask them to not use the SPI jar.

PR #594, opened 2024-01-20
Move OASFactoryResolver to SPI module, bump version to 4.0-SNAPSHOT

MikeEdgar:
For #576


spi/src/main/java/org/eclipse/microprofile/openapi/spi/OASFactoryResolver.java
```diff
@@ -55,7 +48,7 @@ public abstract class OASFactoryResolver {
      * @throws IllegalArgumentException
      *             if an instance could not be created, most likely, due to an illegal or inappropriate class
      */
-    public abstract <T extends Constructible> T createObject(Class<T> clazz);
+    public abstract <T> T createObject(Class<T> clazz);
```
MikeEdgar, line 51: Dropped `extends Constructible` to avoid dependency cycle with `api` module.

api/pom.xml
```diff
@@ -20,7 +20,7 @@
         <!-- This is just for now and will not work if the API has a separate release cycle than the rest. -->
         <groupId>org.eclipse.microprofile.openapi</groupId>
         <artifactId>microprofile-openapi-parent</artifactId>
-        <version>3.2-SNAPSHOT</version>
+        <version>4.0-SNAPSHOT</version>
```
MikeEdgar, line 23: Module version bump required to pass `biz.aQute.bnd:bnd-baseline-maven-plugin` verification.

spi/src/main/java/org/eclipse/microprofile/openapi/spi/OASFactoryResolver.java
MikeEdgar: Implementations should already be checking that `T` is `Constructible` and the API used by applications in `OASFactory` retains the `Constructible` type on `createObject`.

PR #597, opened 2024-01-24
Validate formatting in CI build

Azquelt:
Add a profile to validate code formatting

Use this profile in the github build action

Currently, the build will autoformat the code. This change will ensure that code can't be merged if it wasn't formatted before it was committed (as we did in #580, by making a line too long when updating a comment).


MikeEdgar (approved):
:+1:

Issue #584, opened 2024-01-19
[OAS 3.1.0] Schema changes


Azquelt:
OAS 3.1.0 changes from supporting most of an older JSON schema draft to supporting the whole of the JSON Schema 2020-12 draft [Core][js-core] and [Validation][js-validation].

We need to work out:
- The changes needed to our [`Schema`][schema-model] model class to support the 3.1.0 schema
- Any changes we want to make to our `@Schema` and related annotations in order to allow users to take advantage of the new functionality that is available
- The changes needed to the TCK to expect the new format of schema
  - Existing tests will need changes where JSON schema has made changes to properties
  - Any new functionality will need new tests

Tasks:
- [x] Update [`Schema`][schema-model] model
- [x] Update existing TCKs to expect schemas in the new format
- [x] Write new TCKs which use the new parts of the schema model
- [x] Write new TCKs which test the extensibility of the schema model, particularly in passing through a completely custom schema
- [x] Write new TCKs which test setting custom properties with the same name as properties set by other getters and setters
- [x] Write new TCKs which test setting custom properties to all of the supported types
- [x] Update the `@Schema` annotation to expose new and changed parts of the schema model #601 
- [x] Update TCKs to cover new `@Schema` attributes #601 
- [x] Check whether we've fixed #567

[js-core]: https://json-schema.org/draft/2020-12/json-schema-core
[js-validation]: https://json-schema.org/draft/2020-12/json-schema-validation
[schema-model]: https://github.com/eclipse/microprofile-open-api/blob/3.1.1/api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java

Azquelt:
Here are my initial thoughts:

At a high level, the new JSON schema is very flexible. The core spec defines the idea of a "vocabulary" - a set of properties which have a documented meaning in a JSON schema - and a "dialect" - a set of vocabularies which are supported.

The core spec defines a core vocabulary, and two vocabularies for applying subschemas. The validation spec defines vocabularies for validating the structure and contents of a JSON document, a vocabulary for annotating a schema with metadata like a description and [a dialect][js-dialect] which includes all the required vocabularies of the core and validation specs.

The OpenAPI spec then [defines its own vocabulary][oapi-vocab] containing some extensions, and [a dialect][oapi-dialect] which requires the required JSON schema vocabularies and the OpenAPI vocabulary.

[...]

PR #598, opened 2024-02-06
OAS 3.1 Schema model updates

Azquelt:
For #584


api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -847,18 +893,23 @@ default Schema format(String format) {
...
+     * Updates the type property to either permit or disallow {@code null}
      *
      * @param nullable
-     *            a boolean value indicating this Schema allows a null value.
+     *            a boolean value indicating whether this Schema allows a null value.
+     * @deprecated use {@link #setType(List)}, {@link #addType(SchemaType)}, or {@link #removeType(SchemaType)} to add
+     *             or remove {@link SchemaType#NULL}
      */
+    @Deprecated(since = "4.0")
     void setNullable(Boolean nullable);
```
Azquelt, line 914: I'm not sure whether it's worth keeping these methods or the single argument `setType`.

When I was working on the smallrye implementation, I found that these definitions didn't really retain backwards compatibility because
- `nullable` had three states: `true`, `false` and `null` (not set) whereas the current definition only allows for two. 

  Various parts of the code were checking whether `nullable` had been set by an earlier stage of a process, so the difference between `false` and `null` was important, even though they have the same meaning in the output document.
- it's not clear what `setNullable` should do if `type` is unset.

  If we set `type` to `["null"]` that forbids all types other than `null`, whereas `nullable = true` does nothing if `type` is not set in the old version.
  If we do nothing, then calling `setNullable(true); setType(OBJECT);` leads to a schema where null is not permitted.

It is possible to ensure that `setNullable()` and `setType()` work as they did before, but you need a hidden flag for the value of `nullable` which isn't output and the definition of what these two methods do becomes more complicated so I'm not sure that it's worth it.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -926,29 +926,33 @@ default Schema writeOnly(Boolean writeOnly) {
...
     /**
-     * Returns the example property from this Schema instance.
+     * Returns the first value from the examples property from this Schema instance.
      *
      * @return an object which is an example of an instance of this Schema
-     **/
+     * @deprecated use {@link #getExamples()}
+     */
+    @Deprecated(since = "4.0")
     Object getExample();
```
Azquelt, line 989: `example` is still a valid property [in the OpenAPI spec](https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-19) so I should probably change this back.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -1949,4 +1949,43 @@ default Schema contentSchema(Schema contentSchema) {
...
+     * Sets this schema to a boolean value.
+     * <p>
+     * If this property is not {@code null}, then all other properties are ignored and the schema will be represented by
+     * a boolean {@code true} or {@code false} value.
+     *
+     * @param booleanSchema
+     *            the boolean value of this schema, or {@code null} if it is not a boolean schema
+     * @since 4.0
+     */
+    void setBooleanSchema(Boolean booleanSchema);
```
Azquelt, line 2037: As written, this Javadoc doesn't really describe what other methods do when the schema is a boolean schema. It says "other properties are ignored", which could mean that setting them is ignored, or that they can be set, but they just won't be output.

Reasonable options would be:
- Any other properties set are lost when `setBooleanSchema` is called
  - Calling other setters could either do nothing or throw an exception
  - Calling other getters could either return `null` or throw an exception
- All other properties are kept until the object is serialized, so that `setBooleanSchema(null)` can be called to turn the schema back into an object

Azquelt, line 2037: _Implementation note_: having `setBooleanSchema` made it near-impossible to implement `Schema` as a thin wrapper around a Jackson representation of some JSON.

All the other setters merely change properties on the root object, but to implement this setter you have to change the root object itself. In Jackson, this means setting a new object on the parent and Jackson also doesn't allow you to traverse up a JSON tree (parents know about their children, but children don't know about their parents).

The other way of implementing this would be to have factory methods for an immutable boolean schema on `OASFactory`, but that would be very different than the way the rest of the MP OpenAPI model is designed.

There were other issues with implementing it as a wrapper around Jackson, e.g. calling code may assume that two calls to `getAllOf()` will return the same `Schema` instances, so I gave up on that approach.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -1192,4 +1256,928 @@ default Schema oneOf(List<Schema> oneOf) {
...
+     * BigDecimal minimum = schema.getMinimum(); // returns 3
+     * }
+     * </pre>
+     *
+     * @param propertyName
+     *            the property name
+     * @param the
+     *            value to set
+     */
+    void set(String propertyName, Object value);
```
Azquelt, line 2162: Need to decide on the semantics for `get` and `set` when used with a collection or map.

In general, getters for a collection return a "copy (potentially immutable)", so that you can't get a collection, modify it and have the modification reflected in the model. Setters make no indication of whether they they copy the argument or not. Users are expected to either replace the whole collection, or use the relevant `add` and `remove` methods to modify the collection indirectly.

However, getters returning other model objects do not copy them. Similarly, getters returning `Object` don't copy them, even if the object happens to be a collection.

Given that the primary use case is to get and set arbitrary properties, I think it makes most sense to follow the same semantics for properties where we accept `Object` so I'm leaning towards not requiring a copy here.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
Azquelt: Fixed this.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
Azquelt: Either these need to be removed, or we need TCK tests to assert the specified behavior.

Azquelt: For now I've removed them. I'd be happy to put them back after further discussion on how they should work.

MikeEdgar (approved):
A few comments

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
Azquelt: Do you think this needs to go in the Javadoc or are we happy leaving it vague?

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -1192,4 +1201,928 @@ default Schema oneOf(List<Schema> oneOf) {
...
+
+    /**
+     * Gets all properties of a schema.
+     * <p>
+     * Equivalent to calling {@link #get(String)} for each property and putting them all into a {@code Map}.
+     *
+     * @return a {@code Map} of property names to their corresponding values
+     */
+    Map<String, ?> getAll();
```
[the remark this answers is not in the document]
Azquelt: It would return only those that have been set, matching what is output when the object is serialized.

I hadn't really considered whether we should support keys with a `null` value in the output. The `set` method currently doesn't support setting a `null` value and I haven't seen anything that would need that support in the JSON Schema spec.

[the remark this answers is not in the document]
Azquelt: I've pushed a commit to clarify this

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -1192,4 +1201,929 @@ default Schema oneOf(List<Schema> oneOf) {
...
+     * BigDecimal minimum = schema.getMinimum(); // returns 3
+     * }
+     * </pre>
+     *
+     * @param propertyName
+     *            the property name
+     * @param value
+     *            the value to set, or {@code null} to remove the property
+     */
+    void set(String propertyName, Object value);
```
Azquelt, line 2107: Considering changing this to return the current schema so that it can be chained together when building a schema.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
```diff
@@ -1192,4 +1201,929 @@ default Schema oneOf(List<Schema> oneOf) {
...
+     * <p>
+     * {@code value} must be one of the following types, otherwise non-portable behavior results:
+     * <ul>
+     * <li>Any primitive type
+     * <li>Any primitive wrapper class
+     * <li>{@code null}
+     * <li>{@code String}
+     * <li>{@code BigDecimal}
+     * <li>{@code BigInteger}
+     * <li>Any type which {@link OASFactory} can create
```
Azquelt, line 2086: Noted that in the smallrye implementation, this would require quite a bit of refactoring since the class handling IO for each model type needs to have the IO class for every other model type passed in.

I still think it's a reasonable requirement, but would be open to changing it.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
MikeEdgar: +1

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
MikeEdgar: Is this because of on-the-fly serialization in the smallrye PR?

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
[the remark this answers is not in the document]
MikeEdgar: Thanks, looks good. I am fine with `null` values passed to `set` being removed from the model. As you said, there shouldn't be a place where a `null` value needs to be supported.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
MikeEdgar: I agree that `get` and `set` can be non-copying operations.

api/src/main/java/org/eclipse/microprofile/openapi/models/media/Schema.java
Azquelt: I've made this change.

Azquelt:
I've pushed commits to address the comments above and to add tests for reading and passing through a custom schema dialect via a model reader and a static document.

I think these are all the tests needed for the model parts of #584, so if these are approved I will squash the review change commits and merge this PR.

MikeEdgar (approved):
LGTM

PR #600, opened 2024-03-18
Use the 2.8 microprofile-parent but target Java 11

Azquelt:
Ensures that the TCK uses dependencies from Jakarta EE 9.1.

Keep the tck-bom at 3.2 since the only differences are to use newer versions of libraries which don't support Java 8.


MikeEdgar (approved):
Looks good. My only suggestion is to maybe add a comment to the TCK BOM property stating the reason for the difference from the parent version.

Azquelt:
Yeah, that makes sense. Done.

PR #601, opened 2024-04-16
Schema annotation updates

Azquelt:
Add new fields to `@Schema` and `@SchemaProperty` to support the new properties available in OpenAPI 3.1.

Builds on the model changes from #598

New properties:
- `examples`
- `comment`
- `constValue`
- `ifSchema` / `thenSchema` / `elseSchema`
- `dependentSchemas`
- `contains` / `maxContains` / `minContains`
- `prefixItems`
- `patternProperties`
- `dependentRequired`
- `propertyNames`
- `contentEncoding` / `contentMediaType` / `contentSchema`

Additional changes:
- deprecate `example` in preference of using `examples`
  - keep `example` to preserve backward compatibility
- make `True` and `False` valid everywhere that `Class` is accepted to indicate a subschema
  - previously they were only applicable to `additionalProperties`
- add `additionalProperties` to `@SchemaProperty`
  - I'm unsure why we didn't add it here in 3.1
  - If we don't allow `additionalProperties` in `@SchemaProperty`, then we may want to reconsider adding all the other new properties there also
- Some fixes and clarifications to the `@Schema` javadoc
- Removal of Javadoc indicating restrictions which are no longer present in OpenAPI 3.1 / Json Schema 2020-12

Other notes:
- `type` and `nullable` are unchanged to preserve backward compatibility
- I wanted to add new TCK tests for all the new properties, but I noticed that we don't have tests for the _current_ properties which apply subschemas - we did have some but they were removed in #171 - so I haven't added new ones.


api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
```diff
@@ -40,10 +54,7 @@
 public @interface Schema {
```
MikeEdgar, line 54: What are your thoughts on having a `String value() default ""` property that could be used to set an arbitrary schema (that must be valid JSON/YAML)? That would give support for alternate dialects with the annotation.

I'm not sure myself, but I wanted to mention it as a possibility.

MikeEdgar (approved):
Looks good, thanks @Azquelt

api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/Schema.java
Azquelt: That's certainly an interesting option which I hadn't considered. It becomes a bit more usable since Java 15 added text blocks too.

I'll raise a new issue to think about it.

Azquelt: Opened #610 to continue the discussion after this is merged.

Issue #593, opened 2024-01-19
[OAS 3.1.0] New restrictions on ServerVariable.enum


Azquelt:
Two new restrictions are added to [`ServerVariable.enum`](https://spec.openapis.org/oas/v3.1.0.html#server-variable-object):
- If `enum` is defined it may not be empty
- If `enum` is defined then the value set for `default` must be found within `enum`

We should
- Check that we don't have any tests which violate this constraint
- Update our documentation if necessary

benjamin-confino:
**tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java**

This test modifies the enum list, but never sets a default so the ServerVariable will be in an illegal state. 

(It also raises the question of intermediate values. If you want to remove the defualt from the enum, add a new value, snd set the new value to the default. Is it required to ensure the default is in the enum at every step of the process?)

**The following tests use ServerVariable in line with the rule**

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/JAXRSApp.java - enumeration is unset. 

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ReviewResource.java - enumeration is set, default is set and in the enumeration. 

tck/src/main/java/org/eclipse/microprofile/openapi/reader/MyOASModelReaderImpl.java - enumeration is set, default is set and in the enumeration. 

**The following files only mention ServerVariable**

api/src/main/java/org/eclipse/microprofile/openapi/models/servers/Server.java
api/src/main/java/org/eclipse/microprofile/openapi/models/servers/ServerVariable.java
api/src/main/java/org/eclipse/microprofile/openapi/models/servers/package-info.java
api/src/main/java/org/eclipse/microprofile/openapi/OASFactory.java
api/src/main/java/org/eclipse/microprofile/openapi/annotations/servers/Server.java
api/src/main/java/org/eclipse/microprofile/openapi/annotations/servers/ServerVariable.java
api/src/main/java/org/eclipse/microprofile/openapi/annotations/servers/package-info.java

Azquelt:
> (It also raises the question of intermediate values. If you want to remove the defualt from the enum, add a new value, snd set the new value to the default. Is it required to ensure the default is in the enum at every step of the process?)

No, I don't think this is required, the object being in an invalid intermediate state is fine.

benjamin-confino:
[...]

PR #605, opened 2024-05-01
test methods setDefaultValue and getDefaultValue

benjamin-confino:
When I did the audit for #593 I noticed we don't test setDefaultValue or getDefaultValue so I added a quick check in serverVariableTest


Azquelt:
I don't think this is needed. `processConstructible` called at the start of this method should reflectively test all simple getters and setters.

PR #608, opened 2024-05-02
Fix all checkstyle errors

benjamin-confino:
This fixes all checkstyle errors in the repository

I won't blame you if you want to hold off on this until after https://github.com/eclipse/microprofile-parent/pull/82/files is merged and published to avoid the pom.xml changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.eclipse.microprofile</groupId>
         <artifactId>microprofile-parent</artifactId>
-        <version>2.8</version>
+        <version>2.9-RC1</version>
     </parent>
 
     <groupId>org.eclipse.microprofile.openapi</groupId>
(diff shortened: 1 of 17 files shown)


benjamin-confino:
I've been informed that Microparent will be updated soon, so I'll deliver this later with new version to avoid all that TCK stuff.

pom.xml
```diff
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.eclipse.microprofile</groupId>
         <artifactId>microprofile-parent</artifactId>
-        <version>2.8</version>
+        <version>2.9-RC1</version>
```
Azquelt, line 22: We would need to update this before release.

Azquelt (approved):
I think these changes are fine, as long as we switch to a released version of the microprofile-parent artifact before release. (Either a released 2.9 or back to 2.8 and accept the checkstyle warnings).

pom.xml
Azquelt: Opened #623 so we don't forget to do the update.

Issue #437, opened 2020-07-02
[OAS 3.1.0] PathItem supports references in the Components object.


jmini:
In order to support OpenAPI 3.1.0 (see #333) we need the PathItem to support `$ref`.

> Components Object now has a new entry pathItems, to allow for reusable Path Item Objects to be defined within a valid OpenAPI document.

Reference: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#componentsObject

Example (to be verified):

```yaml
openapi: 3.1.0
info:
  title: Example specification
  version: "1.0"
servers:
- url: http://localhost:8080/
paths:
  /foo:
    $ref: "#/components/pathItems/StandardStatus"
  /bar:
    $ref: "#/components/pathItems/StandardStatus"
components:
  pathItems:
    StandardStatus:
      get:
        parameters:
        - name: extended
          in: query
          schema:
            type: boolean
        responses:
          200:
            description: OK
```

**To be added:**

- [x] The model interface [PathItem.java](https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/models/PathItem.java) is now extending `Reference`.
- [x] The model interface [Components.java](https://github.com/eclipse/microprofile-open-api/blob/master/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Components.java): must manage path items.
- [x] Model TCK tests for new fields
- [x] Model TCK test that `ref` can be set alongside other fields in `PathItem`

Azquelt:
I noticed that both `@CallbackOperation` and the proposed `@Webhook` represent a map entry where the value is a `PathItem`, so these annotations could have a `ref` property added.

Added this to the list in the main comment.

Azquelt:
Thought about this again and I'm less sure. The mapping from annotations to model objects isn't straightforward.

- `@Callback` represents one entry in the map inside the `Callback` object (a pair of `callbackUrlExpression` and `CallbackOperation[]`)
- `@CallbackOperation` represents one entry in the map inside the `PathItem` object (a pair of `method` and `Operation`)

I guess `@CallbackOperation[]` is the part that maps well onto a `PathItem` object and since that's an array rather than an object, it can't have a `ref` property.

You could argue that `@Callback` could have a `ref` to a `PathItem`, but it already has a `ref` property used to reference a `Callback`. It could have a new property used to reference a `PathItem`.

[...]

PR #612, opened 2024-05-10
Add Webhooks model and annotation APIs

### LINKED ISSUE #583: [OAS 3.1.0] Webhooks field in OpenAPI object
OpenAPI 3.1.0 (see #333) adds support for webhooks which the API provider can call, but which may be configured out of band (i.e. in some way other than calling an operation described in the OpenAPI document).

This is similar to the way an Operation can have a callbacks, which represent a call that may be made by the API provider, after a client has called that operation.

Reference: https://spec.openapis.org/oas/v3.1.0.html#openapi-object

[Example](https://github.com/oai/openapi-specification/blob/main/examples/v3.1/webhook-example.yaml):

```yaml
openapi: 3.1.0
info:
  title: Webhook Example
  version: 1.0.0
# Since OAS 3.1.0 the paths element isn't necessary. Now a valid OpenAPI Document can describe only paths, webhooks, or even only reusable components
webhooks:
  # Each webhook needs a name
  newPet:
    # This is a Path Item Object, the only difference is that the request is initiated by the API provider
    post:
      requestBody:
        description: Information about a new pet in the system
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Pet"
      responses:
        "200":
          description: Return a 200 status to indicate that the data was received successfully

components:
  schemas:
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
```

## Work needed

- [x] Add webhooks to the `OpenAPI` model interface
- [x] TCK tests
  - [x] Model reader
  - [x] Static file

Possible extension:
- [ ] Add support for defining webhooks via annotation
  - [ ] TCK tests

Azquelt:
- add `OpenAPI.webhooks`
- add TCKs using model reader and static file
- add `@Webhook` annotation and `@OpenAPIDefinition.webhooks`
- add TCKs using the `@Webhook` annotation

Fixes #583


Azquelt:
I've now updated the list of annotations in the spec.

Azquelt:
Rebased on `main`

api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/Webhook.java
```diff
@@ -0,0 +1,68 @@
...
+ * Describes a named webhook which the API may be instructed to call.
+ * <p>
+ * Webhooks are similar to {@link Callback callbacks} but are set up via an out-of-band mechanism (e.g. an admin web
+ * page) rather than by a call to an operation.
+ *
+ * @since 4.0
+ */
+@Target({})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Webhook {
```
MikeEdgar, line 32: Should this also include `parameters` (applicable to all operations) and `servers` [1] since this is effectively a path item? I wonder also if it should have a `$ref` for the same reason.

[1] https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-6

Azquelt: > Should this also include `parameters` and `servers`

It _could_, but I'm not sure whether we should or not. Path items appear in four places: under `paths`, `callbacks`, `webhooks` and `components` (not yet added).

By far the most common is under `paths` and we don't currently have any way to add `parameters` or `servers` there with annotations since there's nothing in Jakarta REST that really maps to path item and so there's nowhere to put an annotation.

`@Callback` also represents a path item, so if we were to add `parameters` and `servers` to `@Webhook`, we should probably add them to `@Callback` as well.

> I wonder also if it should have a `$ref` for the same reason.

Yes possibly. I was going to think about that under #437.

Another thought occurred to me: if we were to add a new annotation for #437, it would probably have identical fields to `@Webhook` (since both represent one entry in a `Map<String, PathItem>`). We could use one annotation (named `@PathItem`) for both instead.

Azquelt: Let's design this alongside #437.

I'll remove the annotations from this PR so we can get the model parts merged.

Azquelt: Annotations removed.

MikeEdgar (approved):
Looks good

PR #613, opened 2024-05-13
Add mutualTLS to SecuritySchemeType

benjamin-confino:
This resolves #582

PR #615, opened 2024-05-14
590 roles in security schemes

### LINKED ISSUE #590: [OAS 3.1.0] All security schemes may now define required roles
Previously, in a [Security Requirement Object](https://spec.openapis.org/oas/v3.1.0.html#patterned-fields-2), roles were only valid for `oauth2` and `openIdConnect` (where they correspond to required scopes required for execution) but now roles can be specified for all security schemes.

We should:
- Update our documentation to reflect the removal of this limitation
- Add a test which builds a model with roles where they were not previously allowed
- Add a test which uses annotations to declare a security requirement with roles for a previously unsupported type

benjamin-confino:
Resolves #590 

No changes to smallrye were needed to pass the new tests.

Depends on the delivery of #613 - marking this as a draft until 613 is in.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
@@ -49,7 +49,7 @@ public @interface SecurityRequirement {
      * If the security scheme referred by the name property is of type "oauth2" or "openIdConnect", then the scopes
      * array is a list of scope names required for the execution.
      * <p>
-     * For other security scheme types, the array MUST be empty.
+     * For other security scheme types, the array MAY contain a list of required roles.
      * </p>
      *
      * @return a list of scope names required for the execution of this Security Requirement instance.
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
@@ -599,6 +599,7 @@ public class AirlinesAppTest extends AppTestBase {
         String mutualTLS = "components.securitySchemes.mutualTLSScheme.";
         vr.body(mutualTLS + "type", equalTo("mutualTLS"));
         vr.body(mutualTLS + "description", equalTo("mutualTLS authentication needed to manage zepplins"));
+        vr.body(mutualTLS + "scopes", equalTo("zepplinScope"));
     }
 
     @Test(dataProvider = "formatProvider")
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
@@ -549,6 +549,8 @@ public class AirlinesAppTest extends AppTestBase {
                 allOf(aMapWithSize(2),
                         hasEntry(equalTo("userApiKey"), empty()),
                         hasEntry(equalTo("userBearerHttp"), empty()))));
+
+        vr.body("paths.'/zepplins/{id}'.delete.security[0].mutualTLSScheme[0]", equalTo("zepplinScope"));
     }
 
     @Test(dataProvider = "formatProvider")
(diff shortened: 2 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -549,6 +549,8 @@ public void testSecurityRequirement(String type) {
                 allOf(aMapWithSize(2),
                         hasEntry(equalTo("userApiKey"), empty()),
                         hasEntry(equalTo("userBearerHttp"), empty()))));
+
+        vr.body("paths.'/zepplins'/{id}/delete.security.mutualTLSScheme.scopes", equalTo("zepplinScope"));
```
Azquelt, line 553: I think this should be:
```suggestion
        vr.body("paths.'/zepplins/{id}'.delete.security.mutualTLSScheme.scopes", contains("zepplinScope"));
```

It's possible that the quoting and `/` is equivalent, I don't fully understand the query syntax.

However, based on [the spec](https://spec.openapis.org/oas/v3.1.0.html#patterned-fields-2), the value should be a list.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java
```diff
@@ -48,10 +48,9 @@
 
     /**
      * If the security scheme referred by the name property is of type "oauth2" or "openIdConnect", then the scopes
-     * array is a list of scope names required for the execution.
-     * <p>
-     * For other security scheme types, the array MUST be empty.
-     * </p>
+     * array is a list of scope names required for the execution, and the list MAY be empty if authorization does not
+     * require a specified scope. For other security scheme types, the array MAY contain a list of role names which are
+     * required for the execution, but are not otherwise defined or exchanged in-band.
```
Azquelt, line 53: This is copied from the OpenAPI spec (Apache 2.0 license) - need to check if we're ok to include this as-is.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -549,6 +549,8 @@ public void testSecurityRequirement(String type) {
                 allOf(aMapWithSize(2),
                         hasEntry(equalTo("userApiKey"), empty()),
                         hasEntry(equalTo("userBearerHttp"), empty()))));
+
+        vr.body("paths.'/zepplins/{id}'.delete.security.mutualTLSScheme[0][0]", equalTo("zepplinScope"));
```
Azquelt, line 553:
```suggestion
        vr.body("paths.'/zepplins/{id}'.delete.security[0].mutualTLSScheme[0]", equalTo("zepplinScope"));
``

Took me a moment to work out what had happened here. The `security` property contains an array, and so does the `mutualTLSScheme` property. However, since restassured aggregates all properties if you don't specify a list index, you ended up with a list of a list at the end.

I think it's clearer to move the list indices to their actual locations.

Comment (benjamin-confino) [reply tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java:553]: I agree, just noting I got this idiom from many places in the same file so at some point it may be worth cleaning it up

File: api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java (hunk 2 above)
Comment (Azquelt) [reply api/src/main/java/org/eclipse/microprofile/openapi/annotations/security/SecurityRequirement.java:54]: Didn't find this out so I'm going to redo this change, starting from the original javadoc.

File: tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java [hunk 4]
```diff
@@ -460,6 +460,23 @@ public void openAPITest() {
...
+        sr3.addScheme("mutualTLS", "scope");
+        final SecurityRequirement sr4 = createConstructibleInstance(SecurityRequirement.class);
+        sr4.addScheme("HTTP", "scopeTwo");
+        checkSameObject(o, o.addSecurityRequirement(sr3));
+        checkSameObject(o, o.addSecurityRequirement(sr4));
+        assertEquals(o.getSecurity().size(), 2, "The list is expected to contain two entry.");
+        checkListEntry(o.getSecurity(), sr3);
+        checkListEntry(o.getSecurity(), sr4);
+        o.removeSecurityRequirement(sr3);
+        o.removeSecurityRequirement(sr4);
```
Azquelt, line 479: Not sure this test is appropriate here, I think we should move it to the `securityRequirementTest` method in the same class (which currently only adds custom schemes with no scopes).

Azquelt: I've moved this down to `securityRequirementTest`.

Azquelt:

PR #617, opened 2024-05-15
Create a test for an operation without resources

### LINKED ISSUE #585: [OAS 3.1.0] Operation no longer requires responses
It's now valid for an `Operation` not to have a `responses` property.

This is primarily to allow people authoring OpenAPI documents to lay out the operations they need before specifying the responses to them, but we may want to:

* check we don't say it's required in our Javadoc
* add a test which creates a document which has an Operation without responses

benjamin-confino:
This resolves #585

No update to SmallRye was needed to pass these tests.


Azquelt (approved):

Azquelt:

PR #618, opened 2024-05-16
Allow `ref` with description and summary

### LINKED ISSUE #589: [OAS 3.1.0] Summary and description now valid when $ref is set
Previously, when `$ref` was used, nothing else should be set. Now, [Reference objects](https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-18) allow `summary` and `description` to be set along with `$ref`.

We should:
- [x] Update existing documentation for `$ref` #618
- [x] Add tests for setting `summary` and `description` along with `$ref` for objects which support them #618
  - [x] Model tests
  - [x] Annotation tests, where the annotation has `ref` along with `description` and/or `summary`

Azquelt:
`$ref` is now allowed alongside description and summary.

- Update model and annotation javadoc accordingly.
- Test model
- Test annotations

Resolves #589


api/src/main/java/org/eclipse/microprofile/openapi/annotations/links/Link.java
```diff
@@ -97,7 +97,8 @@
      * Reference value to a Link object.
      * <p>
      * This property provides a reference to an object defined elsewhere. This property and all other properties are
-     * mutually exclusive. If other properties are defined in addition to the ref property then the result is undefined.
+     * mutually exclusive. If other properties are defined in addition to the {@code ref} property then the result is
+     * undefined.
```
MikeEdgar, line 101: `Link` may have a `description`

Azquelt: I even used it in the test!

Now fixed, thanks.

MikeEdgar (approved):
LGTM, thank you.

PR #616, opened 2024-05-15
Test for `requestBody` in Operations which don't support a request body

benjamin-confino:
This resolves #591 

This is built on top of #615 and will be marked as a draft until that is delivered.

There is no changes to SmallRye required to pass this test. I did not find any documentation mentioning the removed limitation. And ModelConstructionTest already covers RequestBody

PR #622, opened 2024-05-29
create a model that uses object and space or pipe delimiters

### LINKED ISSUE #586: [OAS 3.1.0] New Parameter style values valid for object type
The [Parameter.style](https://spec.openapis.org/oas/v3.1.0.html#style-values) field specifies the way a parameter is passed. The `spaceDelimited` and `pipeDelimited` values are now valid when `Parameter.type` is `object`.

We may want to:
- Update relevant documentation if we have any
- Add a test which builds a model with these values to ensure they're permitted

benjamin-confino:
Resolves #586

Depends on #616

No changes to smallrye were needed to pass the test


tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
```diff
@@ -1457,6 +1457,22 @@ public void parameterTest() {
...
+        p2.setSchema(schema);
+        p2.setStyle(Parameter.Style.SPACEDELIMITED);
+        assertEquals(p2.getSchema().getType().get(0), Schema.SchemaType.OBJECT);
+        assertEquals(p2.getStyle(), Parameter.Style.SPACEDELIMITED);
+
+        final Parameter p3 = processConstructible(Parameter.class);
+        p3.setSchema(schema);
+        p3.setStyle(Parameter.Style.PIPEDELIMITED);
+        assertEquals(p3.getSchema().getType().get(0), Schema.SchemaType.OBJECT);
+        assertEquals(p3.getStyle(), Parameter.Style.PIPEDELIMITED);
```
Azquelt, line 1475: The `ModelConstructionTest` isn't the right place to test this. That would catch whether the setters immediately reject the value, but given that this restriction affects two parameters, it's much more likely that an implementation would validate it when the model is complete, rather than during construction.

Secondly, this only tests the model and doesn't test that this combination can be configured using annotations.

Azquelt (changes requested):

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
Azquelt: I think to test this we really want a resource method with a parameter in this format.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
```diff
@@ -63,7 +73,11 @@ public Response headZepplin(@RequestBody(ref = "#/paths/~1zepplins~1{id}/delete/
     @Operation(summary = "Deprecate outdated airship technology", operationId = "deprecateZepplin")
     @Produces("text/plain")
     @SecurityRequirement(name = "mutualTLSScheme", scopes = "zepplinScope")
-    public Response getZepplin(@RequestBody(ref = "#/paths/~1zepplins~1{id}/delete/requestBody") String string) {
+    @Parameters(value = {@Parameter(name = "string", description = "something about a string", in = ParameterIn.QUERY,
+                                    required = true,
+                                    schema = @Schema(type = SchemaType.OBJECT), style = ParameterStyle.PIPEDELIMITED)})
```
Azquelt, line 78: `@Parameter` is a repeatable annotation. You can just use it on its own, rather than wrapping it in `@Parameters`.

Azquelt (commented):
This looks better, but I need to do some reading on receiving multipart/form-data requests with Jakarta REST to be sure that the resource methods we're making here are valid.

Azquelt (approved):

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -345,6 +356,21 @@ public void testParameter(String type) {
...
         testUserLoginMethods(vr);
+        testParameterWithObjectAndStyle(vr);
+    }
+
+    private void testParameterWithObjectAndStyle(ValidatableResponse vr) {
+        String headParameters = "paths.'/zepplins/{id}'.head.parameters";
+        String getParameters = "paths.'/zepplins/{id}'.get.parameters";
+
+        vr.body(headParameters, hasSize(1));
+        vr.body(getParameters, hasSize(1));
```
Azquelt, line 367: Hmm, this result is a bit odd. The method also has an `id` path parameter. If this is passing, I would guess that adding `@Parameter` to the method has overriden all the parameters which would have been discovered from the Jakarta REST annotations on the method.

I'm not sure whether the spec requires it to work like that though.

benjamin-confino:
Last push was resolving a merge conflict, no actual changes

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
MikeEdgar: With the `@Path` removed, this now makes sense.

MikeEdgar (approved):

PR #624, opened 2024-05-30
Add tests for encoding with multipart/form-data

### LINKED ISSUE #587: [OAS 3.1.0] More Encoding options can be set for multipart/form-data
The Encoding fields `style`, `explode` and `allowReserved` are now valid when `contentType="multipart/form-data"`.

We may want to:
- Update any documentation which relates to this limitation
  - `@Encoding` annotation
- Add a new test which sets these fields

benjamin-confino:
Fixes #587

Depends on #616

For the comments there was mention of "Shall be ignored if not ... www-form-urlencoded or or multipart/form-data" in several places in the spec, but only one mention of  "Shall be ignored if not ... www-form-urlencoded" in the current comments. I wasn't sure if I should only update the existing one or if I should add new ones too all. I decided there might be a good reason the other methods don't have that, but if you want them all say so and I'll add them.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
@@ -13,20 +13,28 @@
 
 package org.eclipse.microprofile.openapi.apps.airlines.resources;
 
+import java.util.List;
+
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.enums.ParameterIn;
 import org.eclipse.microprofile.openapi.annotations.enums.ParameterStyle;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.enums.SecuritySchemeType;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Encoding;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.media.SchemaProperty;
 import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
 import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirement;
 import org.eclipse.microprofile.openapi.annotations.security.SecurityScheme;
 
+import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.HEAD;
+import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.core.Response;
@@ -69,6 +77,37 @@ public class ZepplinResource {
                required = true,
                schema = @Schema(type = SchemaType.OBJECT), style = ParameterStyle.PIPEDELIMITED)
     public Response getZepplin() {
+        return Response.ok().build();
+    }
+
+    @POST
+    @APIResponse(responseCode = "200", description = "Review posted")
+    @APIResponse(responseCode = "404", description = "Review not found")
+    @Operation(summary = "Post by Zepplin", operationId = "postZepplin")
+    @Produces("text/plain")
+    @SecurityRequirement(name = "mutualTLSScheme", scopes = "zepplinScope")
+    @Consumes("multipart/form-data")
+    public Response postZepplin(
+            @RequestBody(description = "Record of a new user to be created in the system.", required = true,
+                         content = @Content(mediaType = "multipart/form-data",
+                                            schema = @Schema(properties = {
+                                                    @SchemaProperty(name = "passenger", description = "the passanger",
+                                                                    type = SchemaType.STRING),
+                                                    @SchemaProperty(name = "allergies",
+                                                                    description = "Does the passenger have any allergies",
+                                                                    type = SchemaType.ARRAY),
+                                                    @SchemaProperty(name = "specialRequests",
+                                                                    description = "Does the passenger have any special requests",
+                                                                    type = SchemaType.OBJECT)
+                                            }),
+                                            encoding = {
+                                                    @Encoding(name = "allergies",
+                                                              allowReserved = true, explode = true,
+                                                              style = "pipeDelimited"),
+                                                    @Encoding(name = "specialRequests",
+                                                              allowReserved = true, explode = true,
+                                                              style = "spaceDelimited")
+                                            })) List<String> zeeplinUsers) {
         return Response.ok().build();
     }
 }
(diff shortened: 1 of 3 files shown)


Azquelt:
> I wasn't sure if I should only update the existing one or if I should add new ones too all.

The javadoc is a little scatter-shot and inconsistent, so I'm also not sure whether there's a good reason that it's only mentioned in one place. I think your current change that just updates the information where we do have it is the right approach.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
```diff
@@ -58,7 +59,9 @@ public Response headZepplin(@RequestBody(ref = "#/paths/~1zepplins~1{id}/delete/
 
     @GET
     @Path("{id}")
-    @APIResponse(responseCode = "200", description = "Review deleted")
+    @APIResponse(responseCode = "200", description = "Review deleted",
+                 content = @Content(encoding = @Encoding(name = "enigma", style = "pipeDelimited", explode = true,
+                                                         allowReserved = true, contentType = "multipart/form-data")))
```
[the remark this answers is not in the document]
Azquelt: Also, from the [Encoding Object](https://spec.openapis.org/oas/v3.1.0.html#encoding-object) part of the spec:

> `style`: ... This property SHALL be ignored if the **request body media type** is not `application/x-www-form-urlencoded` or `multipart/form-data`.
> _(similar text for `explode` and `allowReserved`)_

I think this means these properties only make sense when applied to the content of a request body where its media type is `multipart/form-data`. (Which makes sense I think, `multipart/form-data` is a type sent by web browsers when the user submits a form).

Note that this is not the `contentType` of the `@Content` annotation, it's the value of `@Consumes` on the method.

Tbh, I'm not that familiar with receiving form data using Jakarta REST, I would probably start by looking up an example and then trying to add the OpenAPI documentation.

Azquelt (changes requested):

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
```diff
@@ -63,4 +73,28 @@ public Response headZepplin() {
...
+    @POST
+    @APIResponse(responseCode = "200", description = "Review posted")
+    @APIResponse(responseCode = "404", description = "Review not found")
+    @Operation(summary = "Post by Zepplin", operationId = "postZepplin")
+    @Produces("text/plain")
+    @SecurityRequirement(name = "mutualTLSScheme", scopes = "zepplinScope")
+    @Consumes("multipart/form-data")
+    public Response postZepplin(
+            @RequestBody(description = "Record of a new user to be created in the system.", required = true,
+                         content = @Content(mediaType = "application/json",
```
MikeEdgar, line 86: I think this should match the `@Consumes` type. Of course consumes could have multiple entries and there could be multiple `@Content` annotations, but for the purposes of the `encoding` they should match.
```suggestion
                         content = @Content(mediaType = "multipart/form-data",
```

MikeEdgar (changes requested):

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
benjamin-confino: All done and tested.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
```diff
@@ -13,20 +13,34 @@
...
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
+=======
 import org.eclipse.microprofile.openapi.annotations.enums.ParameterIn;
 import org.eclipse.microprofile.openapi.annotations.enums.ParameterStyle;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.enums.SecuritySchemeType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+>>>>>>> main
```
Azquelt, line 33: The merge conflicts haven't been resolved here.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
```diff
@@ -71,4 +79,28 @@ public Response headZepplin() {
...
+                                            schema = @Schema(name = "testUser", type = SchemaType.OBJECT,
+                                                             maxProperties = 1024, minProperties = 1, required = true,
+                                                             implementation = User.class),
+                                            encoding = {
+                                                    @Encoding(name = "allergies",
+                                                              allowReserved = true, explode = true,
+                                                              style = "pipeDelimited"),
+                                                    @Encoding(name = "specialRequests",
+                                                              allowReserved = true, explode = true,
+                                                              style = "spaceDelimited")
```
Azquelt, line 102: Every `@Encoding.name` must exist as a named property within the schema in order to satisfy the following requirement from the [Media Type Object](https://spec.openapis.org/oas/v3.1.0.html#media-type-object) encoding field:

> The key, being the property name, MUST exist in the schema as a property.

Azquelt:
I had a look at this from the point of view of whether it's a valid resource method.

As far as I could find, there's no built-in support required for `multipart/form-data` in Jakarta REST 3.0 (EE 9). Support was only required in Jakarta REST 3.1 (EE 10).

However, in that case it would be treated as any other media type without built-in support and the Jakarta REST implementation would [attempt to look up a `MessageBodyReader`][mbr-lookup] and if one was not found, it would generate a `NotSupportedException` which could then be handled by an `ExceptionHandler` to generate a response.

Given this process, I think this method is a valid resource method, even if a compatible implementation may throw a `NotSupportedException` if the method was actually called.

[mbr-lookup]: https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0#message_body_reader

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
Azquelt: It might be better to use a new object to define the schema, rather than `User`, or to use `@Schema.properties` to define the properties needed instead of an implementation object.

tck/src/main/java/org/eclipse/microprofile/openapi/apps/airlines/resources/ZepplinResource.java
```diff
@@ -71,4 +79,32 @@ public Response headZepplin() {
...
+    @Consumes("multipart/form-data")
+    public Response postZepplin(
+            @RequestBody(description = "Record of a new user to be created in the system.", required = true,
+                         content = @Content(mediaType = "multipart/form-data",
+                                            schema = @Schema(properties = {
+                                                    @SchemaProperty(name = "passenger", description = "the passanger"),
+                                                    @SchemaProperty(name = "allergies",
+                                                                    description = "Does the passenger have any allergies"),
+                                                    @SchemaProperty(name = "specialRequests",
+                                                                    description = "Does the passenger have any special requests")
```
Azquelt, line 98: Looks better, can we add set the `type` on each property? `allergies` and `specialRequests` should be `OBJECT`, `passenger` could be whatever.

Azquelt (approved):
Ok, I think this looks good now. Ideally we would squash all the commits.

Azquelt:
Squashed commits and removed merges.

Azquelt:

MikeEdgar (approved):

PR #625, opened 2024-05-30
Test apps with no paths

### LINKED ISSUE #592: [OAS 3.1.0] Only one of Paths, Components, or Webhooks is required
Previously `paths` was always required in an OpenAPI document, but now only one of `paths`, `components` and `webhooks` is required.

In practise, I suspect that usually we will almost always have paths when using MP OpenAPI.

However, we may want to:
- Check and update our documentation in this area
- Add a test building a model without a `paths` entry

benjamin-confino:
fixes #592

No changes to smallrye were required to pass these tests.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppWithJustComponentTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppWithJustComponentTest.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.openapi.tck;
+
+import static org.hamcrest.Matchers.equalTo;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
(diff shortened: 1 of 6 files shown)


tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppWithJustWebHookTest.java
```diff
@@ -0,0 +1,41 @@
...
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, "noPathsAppReader.war")
+                .addPackages(true, "org.eclipse.microprofile.openapi.reader")
+                .addAsManifestResource("microprofile-reader-just-webhook.properties", "microprofile-config.properties");
+    }
+
+    @Test(dataProvider = "formatProvider")
+    public void testDocumentCreated(String type) {
+        ValidatableResponse vr = callEndpoint(type);
+        vr.body("webhooks.blackMarketEvent.get.summary", equalTo("Notifies that a illicit deal has been done"));
```
Azquelt, line 39: Does this pass? It looks like it ought to fail. If it is passing then I'm even more worried.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppWithJustComponentTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import io.restassured.response.ValidatableResponse;
+
+public class ModelReaderAppWithJustComponentTest extends AppTestBase {
+    @Deployment(name = "airlinesModelReader", testable = false)
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, "noPathsAppReader.war")
+                .addPackages(true, "org.eclipse.microprofile.openapi.reader")
+                .addAsManifestResource("microprofile-reader-just-component.properties",
+                        "microprofile-config.properties");
```
Azquelt, line 34: I think this is actually putting the properties file in the wrong place, but the existing tests are also wrong, so I'll raise a new issue for that.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelReaderAppWithJustWebHookTest.java
benjamin-confino: Good news, it needed fixing to pass.

Azquelt (approved):

PR #627, opened 2024-05-31
Make request body required by default

### LINKED ISSUE #349: Make request body required by default
As per @EricWittmann's request following a discussion in https://github.com/smallrye/smallrye-open-api/issues/110#issuecomment-493473681 I propose to make the `required` property of `@RequestBody` true by default
> because more often than not you expect the request body and depend on it

For reference, this is the current state:
https://github.com/eclipse/microprofile-open-api/blob/49e31ea4bf6da48f8491327e1df7f4e47febfc3a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/parameters/RequestBody.java#L59

Azquelt:
When scanning classes and annotations, default to the request body being required for operations which have a request body.

Fixes #349


tck/src/main/java/org/eclipse/microprofile/openapi/apps/petstore/resource/UserResource.java
```diff
@@ -110,9 +109,8 @@ public Response createUsersWithListInput(
     })
     public Response updateUser(
             @Parameter(name = "username", description = "name that need to be deleted",
-                       schema = @Schema(type = SchemaType.STRING),
-                       required = true) @PathParam("username") String username,
```
MikeEdgar, line 114: Should we keep the `required = true` for the `@Parameter`? I know the default is `true` for path parameters anyway, but technically it's a separate change from the linked issue.

Azquelt: Yes, I think that's a mistake.

Azquelt: I've made this change and rebased on `main` to resolve conflicts.

MikeEdgar (approved):
LGTM

MikeEdgar:
@Azquelt ready to merge this one?

Azquelt:
Yep

PR #628, opened 2024-06-04
Use addResource for mp config files

### LINKED ISSUE #626: TCKs put `microprofile-config.properties` in the wrong location
The [MP Config spec states](https://download.eclipse.org/microprofile/microprofile-config-3.1/microprofile-config-spec-3.1.html#default_configsources) properties are read from:

> each property file `META-INF/microprofile-config.properties` found **on the classpath**

However, in some of our tests, we're using `addAsManifestResource` which puts the properties file in `META-INF` in the root of the .war file which is not on the classpath:

https://github.com/eclipse/microprofile-open-api/blob/337f78d394dd15c1cd271c2372fac736fe1b5181/tck/src/main/java/org/eclipse/microprofile/openapi/tck/OASConfigScanDisableTest.java#L33-L38

We should instead use `addAsResource` which ensures the file is placed on the classpath (regardless of the archive type) like this:

```java
     return ShrinkWrap.create(WebArchive.class, "airlines.war") 
             .addPackages(true, "org.eclipse.microprofile.openapi.apps.airlines") 
             .addAsResource("scan-disable-microprofile-config.properties", "META-INF/microprofile-config.properties");
```

benjamin-confino:
This resolves #626 

No changes to smallrye were needed to pass with these changes.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/FilterTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/FilterTest.java
@@ -39,7 +39,7 @@ public class FilterTest extends AppTestBase {
                 .addPackages(true, "org.eclipse.microprofile.openapi.apps.airlines")
                 .addPackages(true, "org.eclipse.microprofile.openapi.filter")
                 .addAsManifestResource("openapi.yaml", "openapi.yaml")
-                .addAsManifestResource("microprofile-config-filter.properties", "microprofile-config.properties");
+                .addAsResource("microprofile-config-filter.properties", "META-INF/microprofile-config.properties");
     }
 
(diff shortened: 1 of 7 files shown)


Azquelt (approved):

Azquelt:
This issue was incorrect, the spec does make this [additional requirement](https://download.eclipse.org/microprofile/microprofile-open-api-3.1.1/microprofile-openapi-spec-3.1.1.html#_location_and_formats):

> For convenience, you may also place your microprofile-config.properties in the root META-INF folder, if you wish to keep both documents in the same directory. This is in addition to the default locations defined by [MicroProfile Config](https://github.com/eclipse/microprofile-config).

PR #629, opened 2024-06-05
PathItem and PathItemOperation annotations

### LINKED ISSUE #437: [OAS 3.1.0] PathItem supports references in the Components object.

### LINKED ISSUE #583: [OAS 3.1.0] Webhooks field in OpenAPI object

### LINKED ISSUE #588: [OAS 3.1.0] Callback.expression can now be a `pathItem` reference as well as an object
A [Callback expression field](https://spec.openapis.org/oas/v3.1.0.html#patterned-fields-1) can now be a reference to a Path Item, as well as a Path Item object directly.

We likely don't need to do anything here since our `PathItem` model object already extends `Reference` (though it probably shouldn't have) but we may want to:
- check if any documentation updates are needed
- add a test which uses a reference here

Note that #437 allows `PathItem`s to be added as components.

Azquelt:
- Allows PathItems to be defined under Components.
- Allows webhooks to be defined.
- Allows Callback to reference a path item.

Fixes #588
Fixes #583 
Fixes #437

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/PathItem.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/PathItem.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.eclipse.microprofile.openapi.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.servers.Server;
+
+/**
+ * Describes a set of operations available on a single path.
+ *
+ * @since 4.0
+ */
+@Target({})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PathItem {
+
+    /**
+     * The name of the Path Item object, used as the map key when the path item is used under
+     * {@link Components#pathItems()} or {@link OpenAPIDefinition#webhooks()}
+     *
+     * @return the path item name
+     */
+    String name();
+
+    /**
+     * Reference value to a PathItem object.
+     * <p>
+     * This property provides a reference to an object defined elsewhere.
+     * <p>
+     * Unlike {@code ref} on most MP OpenAPI annotations, this property is <em>not</em> mutually exclusive with other
+     * properties.
+     *
+     * @return reference to a path item object definition
+     **/
+    String ref() default "";
+
+    /**
+     * The summary of the path item.
+     *
+     * @return the summary
+     */
+    String summary() default "";
+
+    /**
+     * The description of the path item.
+     *
+     * @return the description
+     */
+    String description() default "";
+
+    /**
+     * The operations available under this Path Item.
+     * <p>
+     * The {@link PathItemOperation#method() method} MUST be defined for each operation.
+     *
+     * @return the list of operations
+     */
+    PathItemOperation[] operations() default {};
+
+    /**
+     * A list of servers to be used for this Path Item, overriding those defined for the whole API.
+     *
+     * @return the list of servers
+     */
+    Server[] servers() default {};
+
+    /**
+     * A list of parameters which apply to all operations under this path item.
+     *
+     * @return the list of parameters
+     */
+    Parameter[] parameters() default {};
+
+    /**
+     * List of extensions to be added to the {@link org.eclipse.microprofile.openapi.models.PathItem PathItem} model
+     * corresponding to the containing annotation.
+     *
+     * @return the list of extensions
+     */
+    Extension[] extensions() default {};
+}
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/eclipse/microprofile/openapi/annotations/PathItemOperation.java
```diff
@@ -0,0 +1,175 @@
...
+     **/
+    String method();
+
+    /**
+     * The tag names which apply to this operation.
+     *
+     * @return the list of tag names
+     */
+    String[] tags() default {}; // Could be @Tag[]
+
```
Azquelt, line 56: I remembered there was one bit I wasn't sure about and it was this.

At the moment, we allow `@Tag` to be put on a rest resource method and it will tag the method and create a tag entry at the top level. The user can put a description for the tag in there as well, but if they put a different description in different places then it's undefined which one ends up in the final document.

We could replicate that function here, or we could require the user to just use the tag name, and use the `@Tag` annotation separately if they want a description.

Azquelt: Decided to use the `@Tag` annotation here. It's a bit more verbose but it's consistent with how tags are added to operations elsewhere.

I've done it as a separate commit so it can be easily removed if needed.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/PathItem.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return the path item name
+     */
+    String name();
+
+    /**
+     * Reference value to a PathItem object.
+     * <p>
+     * This property provides a reference to an object defined elsewhere. This property and properties other than
+     * {@link #name()} are mutually exclusive. If properties other than {@code name} are defined in addition to the
+     * {@code ref} property then the result is undefined.
```
MikeEdgar, line 48: Mutual exclusivity between `$ref` and other properties isn't the case for path items, right?

api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/Callback.java
```diff
@@ -77,6 +77,16 @@
...
 
+    /**
+     * Reference value to a Path Item object, to be referenced by the Callback object.
+     * <p>
+     * This property provides a reference to a Path Item object defined elsewhere. {@code #callbackUrlExpression()} is
+     * REQUIRED when this property is set. The referenced Path Item will be used for the URL expression specified.
+     *
+     * @since 4.0
+     */
+    String pathItemRef() default "";
```
MikeEdgar, line 88: It's hard to get my head around a `Callback` being a reference to either another `Callback` or a `PathItem`. Is it correct to say that our `Callback` is effectively a map of `PathItem`s? So, it's basically just one level above a `PathItem`, which is why using the `pathItemRef` requires the `callbackUrlExpression` to complete the "entry".

MikeEdgar (approved):
I think this looks good, thanks. I just have one comment that might require an adjustment and one with me thinking out loud about the `Callback` model :smile:

api/src/main/java/org/eclipse/microprofile/openapi/annotations/PathItemOperation.java
MikeEdgar: +1 to how you've done it.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/PathItem.java
Azquelt: Ah, good spot. Will need a test for this too.

api/src/main/java/org/eclipse/microprofile/openapi/annotations/callbacks/Callback.java
Azquelt: Yep, that's pretty much it.

We could probably do with some examples in the class Javadoc, though we could do with examples on every annotation, along with saying where to use it and what it does.

Azquelt:
Requested changes made and commits squashed.

MikeEdgar (approved):
Looks good, thanks.

PR #644, opened 2024-06-20
Fix spec mistakes found in final readthrough

Azquelt:
Doing a final readthrough of the spec, I noticed a few issues. Mostly minor apart from the Bean Validation requirements which hadn't been updated for OpenAPI 3.1 (though the TCK had been).

- Update example to be compliant with OpenAPI v3.1
- Fix a broken example
- Update Bean Validation requirements for OpenAPI v3.1
- Add schema model methods for custom properties to release notes
- Update source code links to point to the main branch
- Add missing `@since` tags to API classes
- Add myself to the authors list

Forgetting to update the examples is easy to do so I've raised #645 for a future release to make sure we have a TCK for each example and the code is kept in sync.


MikeEdgar (approved):
Looks good, thanks @Azquelt

PR #653, opened 2024-07-31
Update spec license to EFSL v1.1

Azquelt:
License text from https://www.eclipse.org/legal/adoc/efsl-1.1.adoc


Azquelt:
This also needs an updated microprofile-parent with eclipse/microprofile-parent#91 to get the right license in the Javadoc as well.

njr-11 (approved):

Issue #655, opened 2024-08-30
OASFactoryErrorTest TCK test fails when it runs as the first test in the testsuite


TomasHofman:
This is observed in the 3.1 stream, I didn't test the 4.0 stream but I think the issue is present there too.

The OASFactoryErrorTest seems to not be self-sufficient, depends on initialization done by OASScanConfigTests.

This can result in test failures when the tests run in a wrong order. The test order is however random / undefined.

### How the test fails

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   OASFactoryErrorTest.baseInterfaceTest » Test 
Expected exception of type class...
[ERROR]   OASFactoryErrorTest.customAbstractClassTest » Test 
Expected exception of type...
[ERROR]   OASFactoryErrorTest.customClassTest » Test 
Expected exception of type class j...
[ERROR]   OASFactoryErrorTest.extendedBaseInterfaceTest » Test 
Expected exception of ty...
[ERROR]   OASFactoryErrorTest.extendedInterfaceTest » Test 
Expected exception of type c...
[ERROR]   OASFactoryErrorTest.nullValueTest » Test 
Expected exception of type class jav...
[INFO] 
[ERROR] Tests run: 32, Failures: 6, Errors: 0, Skipped: 0
```

The actual exceptions look like:

```
[ERROR] extendedInterfaceTest(org.eclipse.microprofile.openapi.tck.OASFactoryErrorTest)  Time elapsed: 0.023 s  <<< FAILURE!
org.testng.TestException: 

Expected exception of type class java.lang.IllegalArgumentException but got org.testng.TestException: 
Expected exception of type class java.lang.IllegalArgumentException but got java.lang.IllegalStateException: No OASFactoryResolver implementation found!
	at org.testng.internal.ExpectedExceptionsHolder.wrongException(ExpectedExceptionsHolder.java:68)
...
Caused by: java.lang.IllegalStateException: No OASFactoryResolver implementation found!
	at org.eclipse.microprofile.openapi.api@3.1.1//org.eclipse.microprofile.openapi.spi.OASFactoryResolver.instance(OASFactoryResolver.java:82)
	at org.eclipse.microprofile.openapi.api@3.1.1//org.eclipse.microprofile.openapi.OASFactory.createObject(OASFactory.java:96)
[...]

Issue #413, opened 2020-02-26
Why require OpenAPI processing for a non-JAX-RS/Spring deployment?


pferraro:
The following TCK tests:

- ModelConstructionTest
- OASFactoryErrorTest
- StaticDocumentTest

... use web application archives that do not contain any JAX-RS nor Spring applications.  Why should an implementation of MicroProfile OpenAPI be required to perform any OpenAPI processing for such a deployment?

EricWittmann:
I think because it is possible to disable annotation scanning.  Annotation scanning is useful primarily when doing code-first API development.  However, the spec also covers alternative use-cases.  The static document processing support (for example) covers the Design-First approach to API development.  In that case you would design an API (for example using https://studio.apicur.io ) and then implement it using JAX-RS, but configure MP-OpenAPI to include the API design in the deployment and disable annotation scanning.

I suppose the TCK could be made more realistic by actually including some JAX-RS resources in the deployments but configure them to disable scanning.  

Those tests are designed to specifically test the non-annotation scanning functionality of the specification.

kenfinnigan:
From the perspective that MicroProfile will always have JAX-RS present, it's probably more accurate for the tests to include JAX-RS Applications but disable scanning.

jmini:
There is an interesting approach done by the [SmallRye OpenAPI](https://github.com/smallrye/smallrye-open-api) project:

You can use Microprofile OpenAPI on top of Spring…

MikeEdgar:
@pferraro, we discussed this topic in today's OpenAPI call. Do you still view this as an issue in the TCK? Did it cause a problem or is it more of an observation/question?

pferraro:
@MikeEdgar  The side effect of these TCK tests (i.e. those that use an "empty" test application without any JAX-RS endpoints) is that a container must expose an OpenAPI endpoint for *all* web applications, even those that use neither JAX-RS nor Spring, otherwise the container will not pass the MicroProfile OpenAPI TCK.  This does not make sense to me.

MikeEdgar:
We briefly discussed this issue in a recent team call and the suggestion is to update the spec and TCK to indicate that an OpenAPI deployment (i.e. hosting of endpoint `/openapi`) is only required when one or more of the following is true for a deployment/application:

- An OASModelReader has been configured with mp.openapi.model.reader
- An OASFilter has been configured with mp.openapi.filter (maybe?)
[...]

PR #657, opened 2024-09-02
Make ModelConstructionTest, OASFactoryErrorTest activate OpenAPI

TomasHofman:
Proposal to resolve https://github.com/eclipse/microprofile-open-api/issues/655

After reading https://github.com/eclipse/microprofile-open-api/issues/413, I believe these two tests should not be using empty deployment.

OpenAPI should be somehow activated, for the tested functionality to be correctly initialized. The app container have no reason to do this initialization if there is nothing that suggests OpenAPI endpoint is needed.

Currently in Wildfly, these tests only pass if they happen to run after another test from the OpenAPI TCK, where the other test activates OpenAPI for them. They fail if run alone, or if they happen to run as the first test in the module.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ModelConstructionTest.java
@@ -134,7 +134,9 @@ public class ModelConstructionTest extends Arquillian {
 
     @Deployment
     public static WebArchive createDeployment() {
-        return ShrinkWrap.create(WebArchive.class);
+        return ShrinkWrap.create(WebArchive.class)
+                .addPackages(true, "org.eclipse.microprofile.openapi.reader")
+                .addAsManifestResource("microprofile-reader.properties", "microprofile-config.properties");
     }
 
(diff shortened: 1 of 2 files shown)


Azquelt:

MikeEdgar (approved):

Azquelt:
I just want to test this against smallrye and OpenLiberty before I merge it.

TomasHofman:
Thanks for reviewing this!

I would primarily need this in the 3.1 stream. There is currently no branch for 3.1, so I can't open a PR. Is there any chance that 3.1.2 release could be made?

MikeEdgar:
> I would primarily need this in the 3.1 stream. There is currently no branch for 3.1, so I can't open a PR. Is there any chance that 3.1.2 release could be made?

We can probably back-port to 3.1.2. I'll create a branch in history for 3.1.x

TomasHofman:
Thanks! I created the 3.1.x PR.

MikeEdgar (approved):
This change does not cause any issues for smallrye-open-api 4.0.0-prerelease.

Azquelt (approved):
Doesn't cause any issues on OpenLiberty either.

PR #662, opened 2024-10-11
Relax assertions for default header, request body, and response styles

### LINKED ISSUE #661: Several TCK test cases require header `style` set to default value `simple`
There are several TCK test cases that require a header's `style` value is set to `simple`, which is the OAS default for headers [1]. Given that it is the default and is even omitted from the OAS spec examples, it appears this check should be removed from these tests.

A few examples:
- [org.eclipse.microprofile.openapi.tck.FilterTest.testFilterHeader(String)](https://github.com/eclipse/microprofile-open-api/blob/112b7b16806982223d638f415d2979beaf89f6a8/tck/src/main/java/org/eclipse/microprofile/openapi/tck/FilterTest.java#L164)
- [org.eclipse.microprofile.openapi.tck.AirlinesAppTest.testHeaderInAPIResponse(String)](https://github.com/eclipse/microprofile-open-api/blob/112b7b16806982223d638f415d2979beaf89f6a8/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java#L933)
- [org.eclipse.microprofile.openapi.tck.AirlinesAppTest.testHeaderInComponents(String)](https://github.com/eclipse/microprofile-open-api/blob/112b7b16806982223d638f415d2979beaf89f6a8/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java#L991)
- [org.eclipse.microprofile.openapi.tck.AirlinesAppTest.testHeaderInEncoding(String)](https://github.com/eclipse/microprofile-open-api/blob/112b7b16806982223d638f415d2979beaf89f6a8/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java#L958)

[1] https://spec.openapis.org/oas/v3.0.3.html#parameterStyle

MikeEdgar:
Fixes #661

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
@@ -16,6 +16,7 @@
 
 package org.eclipse.microprofile.openapi.tck;
 
+import static org.eclipse.microprofile.openapi.tck.utils.TCKMatchers.hasOptionalEntry;
 import static org.eclipse.microprofile.openapi.tck.utils.TCKMatchers.itemOrSingleton;
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.anEmptyMap;
@@ -408,8 +409,7 @@ public class AirlinesAppTest extends AppTestBase {
         }
 
         bookingParameters = "paths.'/bookings/{id}'.get.parameters";
-        vr.body(bookingParameters + ".findAll { it.name == 'id' }.style", both(hasSize(1)).and(contains("simple")));
-
+        vr.body(bookingParameters + ".findAll { it.name == 'id' }", contains(hasOptionalEntry("style", "simple")));
     }
 
     private void testAvailabilityGetParamater(ValidatableResponse vr) {
@@ -690,7 +690,7 @@ public class AirlinesAppTest extends AppTestBase {
         ValidatableResponse vr = callEndpoint(type);
         String s = "paths.'/user'.post.requestBody.content.'application/json'.encoding.email.";
         vr.body(s + "contentType", equalTo("text/plain"));
-        vr.body(s + "style", equalTo("form"));
+        vr.body(s, hasOptionalEntry("style", "form"));
         vr.body(s + "explode", equalTo(true));
         vr.body(s + "allowReserved", equalTo(true));
         vr.body(s + "x-encoding", equalTo("test-encoding"));
@@ -702,13 +702,13 @@ public class AirlinesAppTest extends AppTestBase {
         String s =
                 "paths.'/user/username/{username}'.put.responses.'200'.content.'application/json'.encoding.password.";
         vr.body(s + "contentType", equalTo("text/plain"));
-        vr.body(s + "style", equalTo("form"));
+        vr.body(s, hasOptionalEntry("style", "form"));
         vr.body(s + "explode", equalTo(true));
         vr.body(s + "allowReserved", equalTo(true));
 
         String t = "paths.'/user/username/{username}'.put.responses.'200'.content.'application/xml'.encoding.password.";
         vr.body(t + "contentType", equalTo("text/plain"));
-        vr.body(t + "style", equalTo("form"));
+        vr.body(t, hasOptionalEntry("style", "form"));
         vr.body(t + "explode", equalTo(true));
         vr.body(t + "allowReserved", equalTo(true));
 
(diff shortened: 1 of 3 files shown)


tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -702,13 +704,13 @@ public void testEncodingResponses(String type) {
         String s =
                 "paths.'/user/username/{username}'.put.responses.'200'.content.'application/json'.encoding.password.";
         vr.body(s + "contentType", equalTo("text/plain"));
-        vr.body(s + "style", equalTo("form"));
+        vr.body(s, hasOptionalEntry("style", "form"));
```
Azquelt, line 707: I'm surprised this works when `s` ends with a `.`

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -702,13 +704,13 @@ public void testEncodingResponses(String type) {
...
         vr.body(s + "explode", equalTo(true));
         vr.body(s + "allowReserved", equalTo(true));
 
         String t = "paths.'/user/username/{username}'.put.responses.'200'.content.'application/xml'.encoding.password.";
         vr.body(t + "contentType", equalTo("text/plain"));
-        vr.body(t + "style", equalTo("form"));
+        vr.body(s, hasOptionalEntry("style", "form"));
```
Azquelt, line 713:
```suggestion
        vr.body(t, hasOptionalEntry("style", "form"));
```

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
```diff
@@ -408,7 +409,8 @@ private void testBookingIdMethods(ValidatableResponse vr) {
         }
 
         bookingParameters = "paths.'/bookings/{id}'.get.parameters";
-        vr.body(bookingParameters + ".findAll { it.name == 'id' }.style", both(hasSize(1)).and(contains("simple")));
+        vr.body(bookingParameters + ".findAll { it.name == 'id' }",
+                both(hasSize(1)).and(hasOptionalEntry("style", "simple")));
```
Azquelt, line 413: Does this not need to be `both(hasSize(1)).and(contains(hasOptionalEntry("style", "simple"))));`?

I think we expect to be matching against a list here.

Azquelt: Actually I think the size check is redundant because `contains` asserts the size.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/utils/TCKMatchers.java
```diff
@@ -169,4 +173,21 @@ private static BigDecimal toBigDecimal(Number number) {
...
+     * @param entryName
+     *            name of the entry in the map
+     * @param validValue
+     *            a single allowed value in the map entry, if present
+     * @return the matcher
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> Matcher<T> hasOptionalEntry(String entryName, Object validValue) {
+        Matcher<T> hasEntry = (Matcher<T>) hasEntry(entryName, validValue);
+        Matcher<T> entryMissing = (Matcher<T>) not(hasKey(entryName));
```
Azquelt, line 189: Hmm, I think this ends up matching anything that isn't a map, as well as a map without the given key, which probably isn't what we want given how difficult it sometimes is to tell what restassured is going to give us to match against.

MikeEdgar: I'll look into that further.

Azquelt: Maybe we need something like `both(isA(Map.class)).and(either(hasEntry).or(entryMissing));`

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
MikeEdgar: Yes, I did some additional debugging on this and indeed the correct object is passed into the matcher.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/AirlinesAppTest.java
MikeEdgar: Changed
```
both(hasSize(1)).and(hasOptionalEntry("style", "simple"))
```
to
```
contains(hasOptionalEntry("style", "simple"))
```

tck/src/main/java/org/eclipse/microprofile/openapi/tck/utils/TCKMatchers.java
MikeEdgar: This is complete. Using `both` did not work so I went with `allOf` instead.

Azquelt (approved):

Azquelt:
This looks good, thank you :)

PR #665, opened 2024-10-16
Remove invalid `content` from TCK simpleapi.yaml test resource

### LINKED ISSUE #663: TCK resource uses invalid type for PathItem response content
The `simpleapi.yaml` resource uses an array for the `content` of a path item response. The correct type is an object. Either we should modify the test resource or define how implementations should handle such a scenario - e.g. the invalid type is treated as `null` and dropped from the input.

https://github.com/eclipse/microprofile-open-api/blob/112b7b16806982223d638f415d2979beaf89f6a8/tck/src/main/resources/simpleapi.yaml#L198-L204

MikeEdgar:
Fixes #663

### DIFF
--- a/tck/src/main/resources/simpleapi.yaml
+++ b/tck/src/main/resources/simpleapi.yaml
@@ -200,7 +200,6 @@ components:
       delete:
         responses:
           '202':
-            content: []
             description: Delete item
       parameters:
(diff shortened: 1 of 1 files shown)


Azquelt (approved):
Thanks

PR #675, opened 2024-12-03
Add json schema dialect property

### LINKED ISSUE #660: `OpenAPI` model object is missing `jsonSchemaDialect` property
We missed the addition of [`jsonSchemaDialect`][openapi-object] property to the OpenAPI object from MP OpenAPI 4.0.

It probably should have been added under #584 which updated the `Schema` object.

[openapi-object]: https://spec.openapis.org/oas/v3.1.0.html#openapi-object

Azquelt:
Add `jsonSchemaDialect` to the model ~and annotations~.

In the model test, I've put the JSON Schema 2020-12 dialect and as it's in the model, I expect it to be passed through as-is.

~In the annotation test, I've put the OAS default schema dialect since I don't think we can reasonably rely on an implementation to have well defined behaviour for any other values. An implementation could reasonably do any of the following:~

- ~generate schemas in the specified dialect~
- ~generate schemas in the OAS default dialect and set `$schema` in all of them, overriding the default set at the top level~
- ~report an error and generate nothing~

Fixes #660

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/models/OpenAPI.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/models/OpenAPI.java
@@ -328,6 +328,36 @@ public interface OpenAPI extends Constructible, Extensible<OpenAPI> {
     void removeWebhook(String name);
 
     /**
+     * Returns the default JSON Schema dialect for schemas in this document.
+     *
+     * @return the identifier of the default schema dialect for schemas in this document
+     * @since 4.1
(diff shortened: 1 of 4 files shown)


MikeEdgar (approved):
This looks good to me. Thanks @Azquelt

Azquelt:
SmallRye passes the TCK with smallrye/smallrye-open-api#2163

MikeEdgar (approved):

PR #678, opened 2025-01-20
Add `Extensible.hasExtension` and `.getExtension`

### LINKED ISSUE #666: Add methods to access a single extension to `Extensible`
Retrieving an extension via `Extensible` currently requires the use of `getExtensions` to acquire the entire extension map, which can then be used to pick the single entry. We should add `getExtension(String)` and `hasExtension(String)` methods to allow implementations to provide more efficient ways to access single entries.

Default methods can be provided to make more efficient implementations optional:

```java
default boolean hasExtension(String name) {
    Map<String, Object> map = getExtensions();
    if (map == null) {
        return false;
    }
    return map.containsKey(name);
}

default Object getExtension(String name) {
    Map<String, Object> map = getExtensions();
    if (map == null) {
        return null;
    }
    return map.get(name);
}
```

Azquelt:
Matching the helper methods for other model maps.

Fixes #666

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/models/Extensible.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/models/Extensible.java
@@ -76,4 +76,34 @@ public interface Extensible<T extends Extensible<T>> {
      */
     void setExtensions(Map<String, Object> extensions);
 
+    /**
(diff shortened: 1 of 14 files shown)


Azquelt:
SmallRye passes the TCK with smallrye/smallrye-open-api#2166

MikeEdgar (approved):
Looks good

PR #679, opened 2025-01-21
Add `@Target` to annotations where it was missing

### LINKED ISSUE #676: `@DependentRequired`, `@DependentSchema` and `@SchemaProperty` are missing `@Target`
I believe all of these annotations are only intended to be used as nested annotations within `@Schema`, so they should have `@Target({})`.

Without `@Target` they can be placed anywhere, but they'll likely be ignored.

Most are also missing `@Retention` so if placed on a Java element, they won't be visible at runtime. When used as a nested annotation (as intended for these annotations), `@Retention` has no effect.

I'm unsure whether fixing this would constitute a breaking change:

- I don't think it would prevent anything loading at runtime (since annotation errors are generally ignored at runtime).
- It would break source-compatibility, users who used these annotations directly on a Java element would get compile errors.
- I don't think any implementation should be processing these annotations if they're placed on a Java element.

Azquelt:
Fixes #676

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/DependentRequired.java
+++ b/api/src/main/java/org/eclipse/microprofile/openapi/annotations/media/DependentRequired.java
@@ -15,6 +15,8 @@
  */
 package org.eclipse.microprofile.openapi.annotations.media;
 
+import java.lang.annotation.Target;
+
 /**
  * A property name and an associated list of other property names.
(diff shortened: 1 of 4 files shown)


Azquelt:
SmallRye still passes all tests with no changes (other than pointing the pom at a snapshot build)

MikeEdgar (approved):

PR #684, opened 2025-02-27
Update dependency versions

Azquelt:
Update the versions of the libraries used by the TCK. Some of them were quite old!


Azquelt:
Smallrye (with the pending implementation changes for 4.1 features) passes the TCK successfully with these updates.

MikeEdgar (approved):
Looks good, thanks @Azquelt

PR #685, opened 2025-03-06
Add dependabot configuration

Azquelt:
Also update the dependencies which the new configuration would otherwise immediately alert us about.


MikeEdgar (approved):
Looks good, we can pick these up in RC2 then.

PR #704, opened 2025-08-21
TCK: Don't assume every Media Type has a schema

### LINKED ISSUE #701: AirlinesOASFilter#filterAPIResponse assumes media type has non-null schema
The `filterAPIResponse` method of `AirlinesOASFilter` assumes that a `MediaType` object has a schema. Because the `schema` of a media type is not required [1], this filter method should be corrected to check the schema's description only if `MediaType#getSchema` returns a non-null value.

This filter class is currently used only by the `FilterTest` test cases.

https://github.com/microprofile/microprofile-open-api/blob/591909d8ccd900513048dc8b590592b2511cf663/tck/src/main/java/org/eclipse/microprofile/openapi/filter/AirlinesOASFilter.java#L127-L135

[1] https://spec.openapis.org/oas/latest.html#considerations-for-file-uploads

Azquelt:
It's valid for a Content object, which represents the content for a Media Type, not to have a schema, so the filter TCK test should not call methods on it without checking whether it's present first.

Fixes #701

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/filter/AirlinesOASFilter.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/filter/AirlinesOASFilter.java
@@ -128,7 +128,7 @@ public class AirlinesOASFilter implements OASFilter {
         if (content != null) {
             if (content.hasMediaType("application/json")) {
                 Schema schema = content.getMediaType("application/json").getSchema();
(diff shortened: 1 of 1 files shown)


Azquelt:
Tested this against https://github.com/smallrye/smallrye-open-api/pull/2324 with the workaround removed.

MikeEdgar (approved):
Looks good, thanks @Azquelt

PR #714, opened 2025-10-19
[issue-713] - Add a test to the TCKs in order to verify the usage of the "@ExternalDocumentation" annotation

### LINKED ISSUE #713: [TCK] - Mising test coverage for the `@ExternalDocumentation` annotation usage
We recently found [an issue in the SmallRye OpenAPI implementation](https://github.com/smallrye/smallrye-open-api/issues/2386), which reveals missing test coverage for the `@ExternalDocumentation` annotation usage in the MicroProfile OpenAPI TCKs.

When such annotation is used at class or method level, the generated OpenAPI document does not contain the expected `externalDocs` property.

fabiobrz:
We recently found [an issue in the SmallRye OpenAPI implementation](https://github.com/smallrye/smallrye-open-api/issues/2386), which reveals missing test coverage for the `@ExternalDocumentation` annotation usage in the MicroProfile OpenAPI TCKs.

This PR provides a new test to cover some basic cases.

Resolves #713

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ExternalDocumentationAnnotationTest.java
+++ b/tck/src/main/java/org/eclipse/microprofile/openapi/tck/ExternalDocumentationAnnotationTest.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.eclipse.microprofile.openapi.tck;
+
+import static org.hamcrest.Matchers.equalTo;
+
+import org.eclipse.microprofile.openapi.apps.scanconfig.ScanConfigApplication;
+import org.eclipse.microprofile.openapi.apps.scanconfig.a.AResource;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.annotations.Test;
+
+import io.restassured.response.ValidatableResponse;
+
+/**
+ * Verify usage of the {@code @ExternalDocumentation} annotation.
+ */
+public class ExternalDocumentationAnnotationTest extends AppTestBase {
+
+    @Deployment(name = "externalDocs-test", testable = false)
+    public static WebArchive deployment() {
+        return ShrinkWrap.create(WebArchive.class, "externalDocs-test.war")
+                .addClasses(ScanConfigApplication.class, AResource.class);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


MikeEdgar:
> I'm a bit surprised we allow `@ExternalDocumentation` to be applied to _any type_ to add documentation at the top level, particularly as we also allow it under `@OpenAPIDefinition`, but that is what the Javadoc currently says (and has said since 1.0).

I was thinking the same thing. For reference, the JavaDoc was added [here](https://github.com/microprofile/microprofile-open-api/pull/31). It's not clear if the use on any type was intentional. I wonder if we should investigate other implementations and if it's not supported anywhere, maybe we alter the JavaDoc accordingly.

fabiobrz:
Thanks @MikeEdgar - I didn't make it to the call, due to conflicts, but I generally agree with your considerations. 
If it can help, while making changes to the WildFly integration in order to support multiple deployments contributing to the OpenAPI documentation, we noticed that annotations reflecting to "centralized" elements (i.e. root elements in the document) can actually be problematic.

fabiobrz:
> We discussed this change on today's spec call and the thinking is that it make sense to go ahead with having the TCK check the use of `@ExternalDocumentation` on REST methods for the 4.2 release of the spec/API.

IIUC, then the test which is checking the method should stay, correct?

> We're generally hesitant about support on any class (which is what the annotation specifies) and we're likely to make some adjustment to that part of the JavaDoc, perhaps limiting it to only REST resource classes. That would need to be in a new major version v5 since it's technically a break in the spec, even if that functionality is not supported by any implementation currently.

Nevertheless, users could object that implementations do not honor what the spec Javadoc states. 
That is the reason why IMHO the test that's checking what happens when a class is annotated should stay as well.

MikeEdgar:
> IIUC, then the test which is checking the method should stay, correct?

Yes, that's right. We would target that change for 4.2 (4.1.1 was just released yesterday).

> Nevertheless, users could object that implementations do not honor what the spec Javadoc states.
> That is the reason why IMHO the test that's checking what happens when a class is annotated should stay as well.

I see your point and it's true that users could object, but I'm dubious that allowing the annotation on any type as stated in the Javadoc was implemented by any runtime or even intended originally. It's not clear what use case it would cover that couldn't be supported using one of several other approaches to having external docs in the root of the produced OpenAPI document. It would probably be best to remove or somehow clarify it in v5 rather than enforcing something that isn't actually going to be used very much.

tck/src/main/java/org/eclipse/microprofile/openapi/tck/ExternalDocumentationAnnotationTest.java
```diff
@@ -0,0 +1,37 @@
...
+     * {@code @ExternalDocumentation} annotation is used on a resource method definition.
+     * @param type Type of the OpenAPI output format
+     */
+    @Test(dataProvider = "formatProvider")
+    public void testExternalDocumentationAnnotationOnMethod(String type) {
+        ValidatableResponse vr = callEndpoint(type);
+
+        vr.body("paths.'/a'.externalDocs.description", equalTo("Find more information about this application resource"));
+        vr.body("paths.'/a'.externalDocs.url", equalTo(
+                "https://github.com/microprofile/microprofile-open-api/blob/main/tck/src/main/java/org/eclipse/microprofile/openapi/apps/scanconfig/a/AResource.java"));
```
Azquelt:
```suggestion
        vr.body("paths.'/a'.get.externalDocs.description", equalTo("Find more information about this application resource"));
        vr.body("paths.'/a'.get.externalDocs.url", equalTo(
                "https://github.com/microprofile/microprofile-open-api/blob/main/tck/src/main/java/org/eclipse/microprofile/openapi/apps/scanconfig/a/AResource.java"));
```

Should be this I think?

`externalDocs` is on the [operation object](https://spec.openapis.org/oas/v3.1.0.html#operation-object), not the [path item object](https://spec.openapis.org/oas/v3.1.0.html#path-item-object).

PR #733, opened 2026-05-20
Specify and add TCK test cases for `@Digits` constraint processing

### LINKED ISSUE #717: Support for `multipleOf` schema assertion from `@Digits` BV constraint
There seems to be an opportunity to specify that the `@Digits` BV contraint results in a `multipleOf` assertion in schemas generated from properties with the annotation. We should investigate adding this to the spec and testing the behavior in the TCK.

Original request in smallrye-open-api: https://github.com/smallrye/smallrye-open-api/issues/2384

MikeEdgar:
Closes #717 

- Add specification of `@Digits` constraint processing for schema properties
- Include new TCK test cases covering the specification
- Fix `TCKMatchers.NUMERIC_COMPARATOR` to create `BigDecimal` instances from string representations of the `Number` arguments, avoiding conversion errors in cast from `float` to `double`.

### DIFF
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
@@ -621,7 +621,7 @@ maxProperties = b`
 | `@DecimalMax(value = a, inclusive = false)` | `number` or `integer` | `exclusiveMaximum = a`
 | `@DecimalMin(value = a)` | `number` or `integer` | `minimum = a`
 | `@DecimalMin(value = a, inclusive = false)` | `number` or `integer` | `exclusiveMinimum = a`
-| `@Digits(integer = <any>, fraction = f)` | `number` or `integer` | `multipleOf` equal to `1` for integer types or 10^-f for non-integer types
+| `@Digits(integer = <any>, fraction = f)` | `number` | `multipleOf = 1` when `fraction` is less than 1, else `multipleOf` equal to 10^-f
 | `@Digits(integer = i, fraction = f)` | `string` | `pattern` matching any string value that satisfies the constraint
 | `@Max(a)` | `number` or `integer` | `maximum = a`
 | `@Min(a)` | `number` or `integer` | `minimum = a`
--- a/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
+++ b/spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
@@ -599,7 +599,7 @@ In some cases, additional schema restrictions can be inferred from Jakarta Bean 
 
 If an implementation includes support for the Jakarta Bean Validation specification, then it must also process Jakarta Bean Validation annotations when creating OpenAPI schemas. Such implementations must add the properties listed in the table below to the schema model when:
 
-* the annotation is applied to to an element for which a schema is generated and
+* the annotation is applied to an element for which a schema is generated and
 * the annotation and generated schema type are listed together in the table below and
 * the annotation has a `group` attribute which is empty or includes `jakarta.validation.groups.Default` and
 * the user has not set any of the relevant property values using other annotations and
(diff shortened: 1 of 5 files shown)


spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
```diff
@@ -611,19 +611,18 @@ If an implementation includes support for the Jakarta Bean Validation specificat
...
-| `@Size(min = a, max = b)` | `object`
-| `minProperties = a +
+| `@Size(min = a, max = b)` | `object` | `minProperties = a +
 maxProperties = b`
 | `@DecimalMax(value = a)` | `number` or `integer` | `maximum = a`
 | `@DecimalMax(value = a, inclusive = false)` | `number` or `integer` | `exclusiveMaximum = a`
 | `@DecimalMin(value = a)` | `number` or `integer` | `minimum = a`
 | `@DecimalMin(value = a, inclusive = false)` | `number` or `integer` | `exclusiveMinimum = a`
+| `@Digits(integer = <any>, fraction = f)` | `number` or `integer` | `multipleOf` equal to `1` for integer types or 10^-f for non-integer types
+| `@Digits(integer = i, fraction = f)` | `string` | `pattern` matching any string value that satisfies the constraint
```
MikeEdgar, line 625: @Azquelt did we discuss having the `string` case in the specification or only the `multipleOf` for numbers?

Azquelt: I'm happy to have both.

The only reason we don't have `@Digits` for strings already was that it was odd to have it for strings but not numbers.

spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
```diff
@@ -611,19 +611,18 @@ If an implementation includes support for the Jakarta Bean Validation specificat
...
 maxItems = b`
-| `@Size(min = a, max = b)` | `object`
-| `minProperties = a +
+| `@Size(min = a, max = b)` | `object` | `minProperties = a +
 maxProperties = b`
 | `@DecimalMax(value = a)` | `number` or `integer` | `maximum = a`
 | `@DecimalMax(value = a, inclusive = false)` | `number` or `integer` | `exclusiveMaximum = a`
 | `@DecimalMin(value = a)` | `number` or `integer` | `minimum = a`
 | `@DecimalMin(value = a, inclusive = false)` | `number` or `integer` | `exclusiveMinimum = a`
+| `@Digits(integer = <any>, fraction = f)` | `number` or `integer` | `multipleOf` equal to `1` for integer types or 10^-f for non-integer types
```
Azquelt: I don't think we need to generate `multipleOf: 1` if the schema type is `integer`, since that's already implied.

MikeEdgar: Maybe a better way to specify this would be in terms of `fraction` rather than in terms of the schema type. When `fraction` < 1, we expect `multipleOf = 1`, otherwise it will be `10^-f`.

spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
MikeEdgar: Thanks, I need to review the BV spec to make sure it's aligned there. I can imagine various gaps in how BV implementations validate cases like `integer = 0` or `fraction = 0`. For example, do we expect `@Digits(integer = 0, fraction = 2)` to allow both `".25"` and `"0.25"` or maybe the TCK just stays away from those edge cases.

spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
Azquelt: Well possibly, but consider this field:

```java
@Digits(integer = 4, fraction = 0)
int myNumber;
```

We'd expect this schema:

```json
{
    "type": "integer",
    "multipleOf": 1
}
```

Here the `multipleOf: 1` is redundant since the schema already requires an integer. I don't think we should require this.

This might mean that in the TCK we have to check for both cases: either the schema type is `integer`, or `multipleOf: 1` is present.

MikeEdgar: The latest commit drops `integer` from the specification, and the TCK test will check `multipleOf = 1` only for `number` types. If the generated schema is `integer` that is acceptable in those cases.

spec/src/main/asciidoc/microprofile-openapi-spec.asciidoc
Azquelt: It wasn't obvious from the javadoc exactly what was required but I looked at one implementation and it appeared to convert to `BigInteger` first so at that point the distinction between `".25"` and `"0.25"` went away. It would have accepted `0.25` for `integer = 0`.

Unless we find somewhere that the spec is more explicit, I think we stay away from those edge cases so that an implementation can match the behaviour of their BV implementation.

Azquelt (approved):
Nice :+1:

PR #739, opened 2026-06-30
Align bnd annotation version with parent

Azquelt:
The parent provides and sets the versions for the BND plugins. Use the same version here where we're adding the annotations as a build-time dependency.


MikeEdgar (approved):
👍

